PR #7, opened 2017-05-17
Add support for armv7

nmorey:
Signed-off-by: Nicolas Morey-Chaisemartin <NMoreyChaisemartin@suse.com>


zoharbmellanox:
Thanks for the patch Nicolas

PR #8, opened 2017-05-23
Fix hang due to polling uninitialized/stale data

DEADBEEFh:
Polling on uninitialized or stale data stored in *poll_buf can lead
to a condition where the latency test server sends a second message
prior to receiving the first message from the client. The second
message causes the client to hang because the value of *poll_buf=2,
but rcnt=1.

This fix sets the value of poll and post buffers to zero prior to
each test iteration.


DEADBEEFh:
Accidentally placed two commits in the same pull request, going to redo as two separate pulls

PR #9, opened 2017-05-24
Fix hang due to polling uninitialized/stale data

DEADBEEFh:
Polling on uninitialized or stale data stored in *poll_buf can lead
to a condition where the latency test server sends a second message
prior to receiving the first message from the client. The second
message causes the client to hang because the value of *poll_buf=2,
but rcnt=1.

This fix sets the value of poll and post buffers to zero prior to
each test iteration.

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1281,6 +1281,10 @@ int create_single_mr(struct pingpong_context *ctx, struct perftest_parameters *u
 		srand(time(NULL));
 		for (i = 0; i < ctx->buff_size; i++) {
 			((char*)ctx->buf[qp_index])[i] = (char)rand();
+			/* WRITE LAT could hang if any buffer value equals 1 */
+			if ((user_param->verb == WRITE) && (user_param->tst == LAT) && ((char*)ctx->buf[qp_index])[i] == 1){
+				((char*)ctx->buf[qp_index])[i] = 0;
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


zoharbmellanox:
please look at the run_iter_lat_write method under src/perftest_resources.c.
i see that poll and post buf pointers are being set to NULL when ever the function is being called.
do we need to set them to zero as well ?

DEADBEEFh:
We shouldn't need to set it there as well since that is just initializing a pointer to NULL, not the value stored there.

However, in reviewing things, I realized my patch fixes one race only to create another.  In the proposed patch, it could be possible for the client to set the *poll_buf to 0 after the server as already sent a message, leading to a new deadlock.

I mistakenly thought the buffer was not initialized, but actually this issue results from the buffer being filled with random data in src/perftest_resources.c function create_single_mr(), but then relying on a specific byte to not equal rcnt++ at the beginning of the server iteration.

I think the cleanest patch might be to either remove the random data fill, or modify the fill to exclude setting any values to 1.

Currently testing a patch that would do the latter and exclude setting the buffer to one if verb=write and test=lat.

DEADBEEFh:
This version of the patch has been running successfully for several days in a loop, so it does resolve the issue, but not sure if you agree if this is the best method or not.

zoharbmellanox:
thanks John , 
we have fix for this issue currently pending for code review,
i will push it ASAP and keep you update

zoharbmellanox:
ive update the Commit that fix the issue,
solution is similar to yours
Thanks for the debug

PR #13, opened 2017-07-27
Correct headers to state the GPL/BSD license

bdrung:
Zohar Ben Aharon confirmed that Mellanox provides perftest to be under GPL/BSD license including the build system files configure.ac and Makefile.am (see linux-rdma mailing list discussion).

Thus adjust the headers of the build system files to clearly state the GPL/BSD license (instead of claiming them to be proprietary).


zoharbmellanox:
Thanks Benjamin.

PR #14, opened 2017-07-27
Debian fixes

bdrung:
Here are the patches that we apply on top of perftest for the official Debian package. Please see the individual commit messages for details.


zoharbmellanox:
Thanks Benjamin for this great commits
i'm sorry for the late response,
can you please provide an example of this hardening flags and the make command that being execution and include them.

bdrung:
Following environment variable is set, when calling ./configure in the Debian package build (example from Debian unstable on amd64):
```
CFLAGS=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security
```

PR #17, opened 2017-11-30
Initialize is_contig_supported for non-experimental verbs

aditr:
Signed-off-by: Adit Ranadive <aditr@vmware.com>


zoharbmellanox:
Thanks For your Commit Adir

PR #19, opened 2017-12-27
Use bash for run_perftest_multi_devices

bdrung:
run_perftest_multi_devices uses bash specific constructs like arrays. Therefore let is use bash instead of sh (which points to dash on Debian).


bdrung:
checkbashisms (from devscripts) is a useful tool to check shell scripts.

PR #21, opened 2018-01-29
Cuda benchmark fix, user control over cuda device

yanivbl6:
- Fixed the missing branch after pp_init_gpu call. I move a section to be encapsulated by {} to avoid future confusion (This results a confusing diff in comparison but this is just a basic encapsulation, and a branch). This fixes gpu-direct benchmark.
- I moved the cuda contexts from being static to the pingpong context, but only if cuda was compiled.
- Added a new parameter for checking which gpu device is being benchmark.

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -3615,6 +3615,7 @@ int run_iter_bw_infinitely(struct pingpong_context *ctx,struct perftest_paramete
 	int 			i = 0;
 	int 			index = 0,ne;
 	int 			err = 0;
+	int			wc_id;
 	#ifdef HAVE_VERBS_EXP
 	struct ibv_exp_send_wr 	*bad_exp_wr = NULL;
 	#endif
@@ -3656,16 +3657,34 @@ int run_iter_bw_infinitely(struct pingpong_context *ctx,struct perftest_paramete
 
 	/* main loop for posting */
 	while (1) {
-
-		/* main loop to run over all the qps and post each time n messages */
+	/* main loop to run over all the qps and post each time n messages */
 		for (index =0 ; index < num_of_qps ; index++) {
 
-			while ((totscnt - totccnt) < user_param->tx_depth) {
+			while ((ctx->scnt[index] - ctx->ccnt[index]) < user_param->tx_depth) {
 				if (ctx->send_rcredit) {
 					uint32_t swindow = scnt_for_qp[index] + user_param->post_list - ctx->credit_buf[index];
 					if (swindow >= user_param->rx_depth)
 						break;
 				}
+
+				if (user_param->post_list == 1 && (ctx->scnt[index] % user_param->cq_mod == 0 && user_param->cq_mod > 1)) {
+
+					#ifdef HAVE_VERBS_EXP
+					#ifdef HAVE_ACCL_VERBS
+					if (user_param->verb_type == ACCL_INTF)
+						ctx->exp_wr[index].exp_send_flags &= ~IBV_EXP_QP_BURST_SIGNALED;
+					else {
+					#endif
+						if (user_param->use_exp == 1)
+							ctx->exp_wr[index].exp_send_flags &= ~IBV_EXP_SEND_SIGNALED;
+						else
+					#endif
+							ctx->wr[index].send_flags &= ~IBV_SEND_SIGNALED;
+					#ifdef HAVE_ACCL_VERBS
+					}
+					#endif
+				}
+
 				#ifdef HAVE_VERBS_EXP
 				if (user_param->use_exp == 1)
 					err = (ctx->exp_post_send_func_pointer)(ctx->qp[index],&ctx->exp_wr[index*user_param->post_list],&bad_exp_wr);
@@ -3682,6 +3701,26 @@ int run_iter_bw_infinitely(struct pingpong_context *ctx,struct perftest_paramete
 				ctx->scnt[index] += user_param->post_list;
 				scnt_for_qp[index] += user_param->post_list;
 				totscnt += user_param->post_list;
+
+				/* ask for completion on this wr */
+				if (user_param->post_list == 1 &&
+						(ctx->scnt[index]%user_param->cq_mod == user_param->cq_mod - 1 ||
+							(user_param->test_type == ITERATIONS && ctx->scnt[index] == user_param->iters - 1))) {
+					#ifdef HAVE_VERBS_EXP
+					#ifdef HAVE_ACCL_VERBS
+					if (user_param->verb_type == ACCL_INTF)
+						ctx->exp_wr[index].exp_send_flags |= IBV_EXP_QP_BURST_SIGNALED;
+					else {
+					#endif
+						if (user_param->use_exp == 1)
+							ctx->exp_wr[index].exp_send_flags |= IBV_EXP_SEND_SIGNALED;
+						else
+					#endif
+							ctx->wr[index].send_flags |= IBV_SEND_SIGNALED;
+					#ifdef HAVE_ACCL_VERBS
+					}
+					#endif
+				}
 			}
 		}
 		if (totccnt < totscnt) {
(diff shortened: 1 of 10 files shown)


yanivbl6:
For example, in a system when device 0 is on the same socket as the Nic but device 1 is on different socket, but all paths pass through Intel root complex:

 ./ib_send_bw 10.143.119.27 -p 10000 --report_gbit --cuda_device 0
 #bytes     #iterations    BW peak[Gb/sec]    BW average[Gb/sec]   MsgRate[Mpps]
 65536      1000             31.64              30.02              0.057259

./ib_send_bw 10.143.119.27 -p 10000 --report_gbit --cuda_device 1
 #bytes     #iterations    BW peak[Gb/sec]    BW average[Gb/sec]   MsgRate[Mpps]
 65536      1000             10.08              10.06              0.019190

This also apply for the receiving side:
./ib_send_bw -p 10000 --report_gbit --cuda_device 0
 #bytes     #iterations    BW peak[Gb/sec]    BW average[Gb/sec]   MsgRate[Mpps]
 65536      1000             0.00               82.19              0.156769

./ib_send_bw -p 10000 --report_gbit --cuda_device 1
 #bytes     #iterations    BW peak[Gb/sec]    BW average[Gb/sec]   MsgRate[Mpps]
 65536      1000             0.00               2.49               0.004754

zoharbmellanox:
Hi Yaniv ,
Thanks for your commit and sorry for the delay.
could you merge this two commits to single ? 
also after change the deviceID into deviceCount please change all variables that have device ID to device count.

yanivbl6:
Combining the commits at this stage would be bothersome (Require force push/ new PR), is _squash and merge_ not possible?
I didn't understand the changing deviceID into deviceCount comment, Can you comment on the code?
DeviceID should not be equal to deviceCount, it should always be lower to represent an available device.

src/perftest_resources.c
```diff
@@ -80,28 +78,30 @@ static int pp_init_gpu(struct pingpong_context *ctx, size_t _size)
 		return 1;
 	} else if (deviceCount == 1)
 		printf("There is 1 device supporting CUDA\n");
-	else
-		printf("There are %d devices supporting CUDA, picking first...\n", deviceCount);
-
-	int devID = 0;
+	else if (deviceCount > devID)
```
zoharbmellanox: here you compare deviceCount with deviceID,  could you explain if this reamingin from this commit when you refer deviceID as deviceCount? 
please explain HL what is the meaning of this testing ?

yanivbl6: I don't remember refering deviceID as deviceCount. 

The second commit it is just fixing the description to to the --help output for the new parameters and fixes an error with the parser routine where the parser did not expect a second argument after the the --cuda_device

The comparison is required to make sure the user specified an recognized device. If I use a server with 2 GPUs I can pick device 0 or 1 but device 2 would result an error. Nothing fancy.

zoharbmellanox: could there be situation where we support more the 2 GPU?  
please add the possibility also of 0 GPU's.

yanivbl6: 0 GPUs (assuming Cuda is available) will result the error:
Error- There are 0 devices supporting CUDA, can not pick device %d

Regarding more than two GPUs- I don't see why not.

yanivbl6:
BTW, One can also use cuda environment variables to specify GPU, without this patch. I think this way is better, but there are issues with it please merge Idit's fix, as fixing the broken Cuda test is several priorities aboves (It is currently bluffing regarding the performance). The rest of what I did is mostly cosmetic and I will not have time to fix it in the immediate future.

byronyi:
@yanivbl6 Hi Yaniv, is GPUDirect feature in perftest usable? I have taken a shot and got the following error:

```
$ ~/perftest/ib_write_bw -d mlx5_2 --use_cuda

************************************
* Waiting for client to connect... *
************************************
initializing CUDA
There are 4 devices supporting CUDA, picking first...
[pid = 7722, dev = 0] device name = [Tesla V100-SXM2-32GB]
creating CUDA Ctx
making it the current CUDA Ctx
cuMemAlloc() of a 131072 bytes GPU buffer
allocated GPU buffer address at 000001001e000000 pointer=0x1001e000000
                    RDMA_Write BW Test
 Dual-port       : OFF		Device         : mlx5_2
 Number of qps   : 1		Transport type : IB
 Connection type : RC		Using SRQ      : OFF
 CQ Moderation   : 100
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 3
 Max inline data : 0[B]
 rdma_cm QPs	 : OFF
 Data ex. method : Ethernet
 local address: LID 0000 QPN 0x00e2 PSN 0x7f3b1d RKey 0x00773c VAddr 0x0001001e010000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:10:00:21:03
 remote address: LID 0000 QPN 0x00e2 PSN 0x14accf RKey 0x0041ac VAddr 0x0001001e010000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:10:00:27:02
 #bytes     #iterations    BW peak[MB/sec]    BW average[MB/sec]   MsgRate[Mpps]
ethernet_read_keys: Couldn't read remote address
 Unable to read to socket/rdam_cm
 Failed to exchange data between server and clients
```

I am not sure if there is any configuration error, or should I use a different version of CUDA/nv_peer_mem/perftest. Have you ever run this test recently? What is your recommended software configuration?

yanivbl6:
I ran the mofed version not too long ago, and it worked. I think it was with Infiniband though, and I am getting the sense that this is Ethernet related. Can you open an issue?

byronyi:
On the other side I got:

```
$  ~/perftest/ib_write_bw -d mlx5_2 --use_cuda 172.16.0.76
initializing CUDA
There are 4 devices supporting CUDA, picking first...
[pid = 23817, dev = 0] device name = [Tesla V100-SXM2-32GB]
creating CUDA Ctx
making it the current CUDA Ctx
cuMemAlloc() of a 131072 bytes GPU buffer
allocated GPU buffer address at 000001001e000000 pointer=0x1001e000000
                    RDMA_Write BW Test
 Dual-port       : OFF		Device         : mlx5_2
 Number of qps   : 1		Transport type : IB
 Connection type : RC		Using SRQ      : OFF
 TX depth        : 128
 CQ Moderation   : 100
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 3
 Max inline data : 0[B]
 rdma_cm QPs	 : OFF
 Data ex. method : Ethernet
 local address: LID 0000 QPN 0x0135 PSN 0xcb3b11 RKey 0x00c9ce VAddr 0x0001001e010000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:10:00:21:03
 remote address: LID 0000 QPN 0x0135 PSN 0x74aad5 RKey 0x00b96b VAddr 0x0001001e010000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:10:00:27:02
 #bytes     #iterations    BW peak[MB/sec]    BW average[MB/sec]   MsgRate[Mpps]
mlx5: right-wombat: got completion with error:
00000000 00000000 00000000 00000000
00000000 00000000 00000000 00000000
00000000 00000000 00000000 00000000
00000000 02005104 08000135 000046d2
 Completion with error at client
 Failed status 4: wr_id 0 syndrom 0x51
scnt=128, ccnt=0
 Failed to complete run_iter_bw function successfully
```

I ran the test successfully by only using ``--with_cuda`` on one side. 

The working combinations are:

```
server$ ib_write_bw --use_cuda 
client$ ib_write_bw <server_ip>
```

```
server$ ib_read_bw
client$ ib_read_bw --use_cuda <server_ip>
```

```
server$ ib_send_bw --use_cuda 
client$ ib_send_bw <server_ip>
```

I guess it is related to GPUDirect RDMA READ.

yanivbl6:
That's a weird error. Can you try increasing the tx-depth (-t flag) and see if it solves anything?
try increasing rx-depth to fit, I don't know if it's ok to have bigger tx than rx.

I will try replicating it next week, but I don't have available GPU servers with RoCE at the moment.

yanivbl6:
I tried replicating the error, building perftest from source (master) with both IB and RoCE, and it worked fine in both directions.

byronyi:
Last time I tried with CUDA 9.0 and NVIDIA 384.145. We are upgrading to CUDA 10 and NVIDIA 410.67, and I will try again.

Btw, did you see visible performance gap between GPUDirect READ and WRITE? Some claim the performance of READ is a lot more worse than WRITE, and we are trying to look into that.

yanivbl6:
I am not aware of that, checking it on an Intel GPU-direct "unfriendly" server, I got the same bandwidth for both read and write. @bureddy, are you maybe familiar with this?

drossetti:
Available GPU RDMA BW varies depending on the PCIE architecture (PEX vs IOH), GPU Model, IOMMU and ACS, etc.
An old post on the subject is https://devblogs.nvidia.com/benchmarking-gpudirect-rdma-on-modern-server-platforms/

PR #29, opened 2018-03-14
Set default un-support for contig memory

zoharbmellanox:
By default we will not support this since it is EXP verbs supported

Signed-off-by: Zohar Ben Aharon <zoharb@mellanox.com>


zoharbmellanox:
Fix the support for contig memory

larrystevenwise:
I tested this out and it resolves the bug.

PR #35, opened 2018-05-14
Patches related to inline data and byte unit prints

tatyana-en:
perftest: Use the unambiguous byte unit MiB instead of MB
    
perftest: Use no greater than the max supported inline data size
    
perftest: Add Intel devices to the perftest device list

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -1487,6 +1487,8 @@ enum ctx_device ib_dev_name(struct ibv_context *context)
 		/* Assuming it's Mellanox HCA or unknown.
 		If you want Inline support in other vendor devices, please send patch to gilr@dev.mellanox.co.il
 		*/
+	} else if (attr.vendor_id == 0x8086) {
+		dev_fname = INTEL_ALL;
 	} else {
 
 		switch (attr.vendor_part_id) {
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -432,9 +432,9 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 		printf(" Report RX & TX results separately on Bidirectinal BW tests\n");
 
 		printf("      --report_gbits ");
-		printf(" Report Max/Average BW of test in Gbit/sec (instead of MB/sec)\n");
-		printf("        Note: MB=2^20 byte, while Gb=10^9 bits. Use these formulas for conversion:\n");
-		printf("        Factor=10^9/(20^2*8)=119.2; MB=Gb_result * factor; Gb=MB_result / factor\n");
+		printf(" Report Max/Average BW of test in Gbit/sec (instead of MiB/sec)\n");
+		printf("        Note: MiB=2^20 byte, while Gb=10^9 bits. Use these formulas for conversion:\n");
+		printf("        Factor=10^9/(20^2*8)=119.2; MiB=Gb_result * factor; Gb=MB_result / factor\n");
 
 		if (connection_type != RawEth) {
 			printf("      --report-per-port ");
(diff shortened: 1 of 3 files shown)


src/perftest_parameters.c
```diff
@@ -432,9 +432,9 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 		printf(" Report RX & TX results separately on Bidirectinal BW tests\n");
 
 		printf("      --report_gbits ");
-		printf(" Report Max/Average BW of test in Gbit/sec (instead of MB/sec)\n");
-		printf("        Note: MB=2^20 byte, while Gb=10^9 bits. Use these formulas for conversion:\n");
-		printf("        Factor=10^9/(20^2*8)=119.2; MB=Gb_result * factor; Gb=MB_result / factor\n");
+		printf(" Report Max/Average BW of test in Gbit/sec (instead of MiB/sec)\n");
```
xXxSiMoNxXx: I agree that MiB the correct term to use here, but it is not widely used and I am afraid it will lead to misunderstanding here. Lets see what other people think.

src/perftest_parameters.c
```diff
@@ -495,7 +495,7 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 		printf(" Set the maximum rate of sent packages. default unit is [Gbps]. use --rate_units to change that.\n");
 
 		printf("      --rate_units=<units>");
-		printf(" [Mgp] Set the units for rate limit to MBps (M), Gbps (g) or pps (p). default is Gbps (g).\n");
```
xXxSiMoNxXx: Same here

src/perftest_parameters.c
amirancel: I agree

src/perftest_resources.c
```diff
@@ -1836,6 +1836,9 @@ struct ibv_qp* ctx_exp_qp_create(struct pingpong_context *ctx,
 		printf("  Actual inline-receive(%d) < requested inline-receive(%d)\n",
 				attr.max_inl_recv, user_param->inline_recv_size);
 
+	if (user_param->inline_size > attr.cap.max_inline_data)
```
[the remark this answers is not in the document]
tatyana-en: The messages above are referring to different error checks so they shouldn't be changed.
This one is for:  if (user_param->inline_recv_size > attr.max_inl_recv)
I can add a debug print for the new check, if you prefer.

src/perftest_resources.c
```diff
@@ -1889,6 +1892,9 @@ struct ibv_qp* ctx_qp_create(struct pingpong_context *ctx,
 		fprintf(stderr, "Requested SQ size might be too big. Try reducing TX depth and/or inline size.\n");
 		fprintf(stderr, "Current TX depth is %d and  inline size is %d .\n", user_param->tx_depth, user_param->inline_size);
 
+	if (user_param->inline_size > attr.cap.max_inline_data)
```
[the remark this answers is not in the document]
tatyana-en: The messages above are referring to different error checks so they shouldn't be changed.
This one is for:  if (errno == ENOMEM)
I can add a debug print for the new check, if you prefer.

src/perftest_parameters.c
tatyana-en: I agree that MiB is less known, however MB is well known and people are wondering why they are getting very different BW numbers in MB units while using different tools, including perftest. The motivation behind this change is to make it less confusing. Unless your users run only benchmarks which generate results in MiB while reporting them in MB, your users are also going to be confused by this.

src/perftest_resources.c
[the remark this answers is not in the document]
xXxSiMoNxXx: Yes, please.
Those type of changes are not visible to well to the user.
I think it should be visible, since it affects the performance.

src/perftest_resources.c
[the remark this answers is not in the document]
xXxSiMoNxXx: Same as above.

src/perftest_parameters.c
xXxSiMoNxXx: I prefer to leave it as it was.

PR #38, opened 2018-08-27
fix netinet/ip.h includes

grafoo:
do not use absolute path of header file in include directives.


dimasique1:
same fix is part of https://github.com/linux-rdma/perftest/pull/48, abandon the PR

PR #40, opened 2018-10-16
Use GCC built-in atomic function

zorrohahaha:
Description:
    When testing ib_write_bw with "--run_infinitely", the function
	catch_alarm_infintely in handle_signal_print_thread thread will set
	the iterations to zero. But in aarch64 platform, it is unsafe. The
	working thread will overwrite the 0 to the number loaded previously,
	so the action set to zero doesn't take effect. And we could see that
	sometimes the result is double or even triple of the average/correct
	value. Using the atomic add function instead of the "+" operator will
	get rid of the risk.

Signed-off-by: Bing Zhao <ilovethull@163.com>

### DIFF
--- a/src/perftest_parameters.h
+++ b/src/perftest_parameters.h
@@ -356,7 +356,7 @@ struct perftest_parameters {
 	enum ibv_mtu			curr_mtu;
 	uint64_t			size;
 	uint64_t			dct_key;
-	int				iters;
+	volatile int			iters;
 	uint64_t			iters_per_port[2];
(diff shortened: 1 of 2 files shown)


dimasique1:
Will be able to merge it when set compilation tests on all supported platforms. Or you need to add define for aarch64, so it is 100% doesn't influence other platforms

dimasique1:
I am closing the PR, because the issue seems to be solved in rdma-core-dc-support, which planned to become the main branch soon. If the issue persists, please, resend it. Thanks!

PR #48, opened 2019-01-31
Patches from Debian package

bdrung:
This pull request contains the two patches from the Debian package. See the individual commits for details.


rleon:
Why wasn't this merged?

PR #58, opened 2019-07-21
Allow overriding CQ moderation on post list mode

firasj:
When posting list of WQEs, allow specifying CQ moderation instead of
overriding it.

For that, request SIGNAL every X WQEs, and fix BW tests to support the
case that cq_mod != post_list.

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -2082,8 +2082,8 @@ int parser(struct perftest_parameters *user_param,char *argv[], int argc)
 					  return 1;
 				  } break;
 			case 'Q': CHECK_VALUE(user_param->cq_mod,int,MIN_CQ_MOD,MAX_CQ_MOD,"CQ moderation");
-					  user_param->req_cq_mod = 1;
-					  break;
+				  user_param->req_cq_mod = 1;
+				  break;
 			case 'A':
 				  if (user_param->verb != ATOMIC) {
 					  fprintf(stderr," You are not running the atomic_lat/bw test!\n");
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -2916,14 +2916,20 @@ void print_report_bw (struct perftest_parameters *user_param, struct bw_report_d
 
 	if (user_param->noPeak == OFF) {
 		/* Find the peak bandwidth unless asked not to in command line */
-		for (i = 0; i < num_of_calculated_iters * num_of_qps; ++i) {
-			for (j = i; j < num_of_calculated_iters * num_of_qps; ++j) {
+		for (i = 0; i < num_of_calculated_iters * num_of_qps; i += user_param->post_list) {
+			for (j = ROUND_UP(i + 1, user_param->cq_mod) - 1; j < num_of_calculated_iters * num_of_qps;
+					j += user_param->cq_mod) {
 				t = (user_param->tcompleted[j] - user_param->tposted[i]) / (j - i + 1);
-				if (t < opt_delta) {
+				if (t < opt_delta)
 					opt_delta  = t;
-					opt_posted = i;
-					opt_completed = j;
-				}
+			}
+
+			/* Handle case where CQE was explicitly signaled on last iteration. */
+			if ((num_of_calculated_iters * num_of_qps) % user_param->cq_mod) {
+				j = num_of_calculated_iters * num_of_qps - 1;
+				t = (user_param->tcompleted[j] - user_param->tposted[i]) / (j - i + 1);
+				if (t < opt_delta)
+					opt_delta  = t;
 			}
 		}
 	}
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -647,6 +647,7 @@ static void init_perftest_params(struct perftest_parameters *user_param)
 	user_param->rx_depth		= user_param->verb == SEND ? DEF_RX_SEND : DEF_RX_RDMA;
 	user_param->duplex		= OFF;
 	user_param->noPeak		= OFF;
+	user_param->req_cq_mod		= 0;
 	user_param->cq_mod		= DEF_CQ_MOD;
 	user_param->iters		= (user_param->tst == BW && user_param->verb == WRITE) ? DEF_ITERS_WB : DEF_ITERS;
 	user_param->dualport		= OFF;
@@ -975,9 +976,15 @@ static void force_dependecies(struct perftest_parameters *user_param)
 	}
 
 	if (user_param->post_list > 1) {
-		user_param->cq_mod = user_param->post_list;
-		printf(RESULT_LINE);
-		printf("Post List requested - CQ moderation will be the size of the post list\n");
+		if (!user_param->req_cq_mod) {
+			user_param->cq_mod = user_param->post_list;
+			printf(RESULT_LINE);
+			printf("Post List requested - CQ moderation will be the size of the post list\n");
+		} else if ((user_param->post_list % user_param->cq_mod) != 0) {
+			printf(RESULT_LINE);
+			fprintf(stderr, " Post list size must be a multiple of CQ moderation\n");
+			exit(1);
+		}
 	}
 
 	if (user_param->test_type==DURATION) {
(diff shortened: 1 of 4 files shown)


src/perftest_parameters.c
```diff
@@ -961,9 +961,15 @@ static void force_dependecies(struct perftest_parameters *user_param)
...
 
 	if (user_param->post_list > 1) {
-		user_param->cq_mod = user_param->post_list;
-		printf(RESULT_LINE);
-		printf("Post List requested - CQ moderation will be the size of the post list\n");
+		if (user_param->cq_mod == DEF_CQ_MOD) {
+			user_param->cq_mod = user_param->post_list;
+			printf(RESULT_LINE);
+			printf("Post List requested - CQ moderation will be the size of the post list\n");
+		} else if ((user_param->post_list % user_param->cq_mod) != 0) {
```
dimasique1: Why must post list size be a multiple of CQ moderation? 
Note that in default case CQ mod = 100, post_list = 1, but 1 % 100 != 0 and it is just fine.

CQ moderation is how often we request a CQE per WQE, post_list is how often we ring the doorbell per WQE.

src/perftest_resources.c
```diff
@@ -3253,12 +3253,14 @@ void ctx_set_send_reg_wqes(struct pingpong_context *ctx,
...
-				ctx->wr[i*user_param->post_list + j].send_flags = IBV_SEND_SIGNALED;
 				ctx->wr[i*user_param->post_list + j].next = NULL;
+			} else {
+				ctx->wr[i*user_param->post_list + j].next = &ctx->wr[i*user_param->post_list+j+1];
 			}
 
-			else {
-				ctx->wr[i*user_param->post_list + j].next = &ctx->wr[i*user_param->post_list+j+1];
+			if ((j + 1) % user_param->cq_mod == 0) {
+				ctx->wr[i*user_param->post_list + j].send_flags = IBV_SEND_SIGNALED;
```
dimasique1: Taking into consideration the comment above, the check should be somewhat like ((i*user_param->post_list + j + 1) % user_param->cq_mod == 0).

src/perftest_parameters.c
firasj: Previous design was to request a CQE on last WQE of post list.
My original proposal is to be able to request multiple CQEs per post list, but for that to work, we need a CQE on the last WQE of the post list. Otherwise, the test might get stuck if last posted list doesn't end with a CQE.

But on second thought after your comment, I think the condition should be that #iterations is a multiple of CQ mod to guarantee no hangs. Do you prefer to have this check explicitly or same as post list?

From README/Known Issues:
>  8. for postlist, tests will not finish correctly when using iterations and num_iters/poslist_val has a remainder.

dimasique1: Agreed with #iterations is a multiple of CQ mod and a multiple of postlist_val. Explicit checks and error messages are better from my viewpoint.

> the test might get stuck if last posted list doesn't end with a CQE.
it was not expected by me, are you sure? Cause if postlist_val == 1 we are good with any CQE mod. What should be the difference with postlist_val == 2 and CQE mod == 100?

firasj: Are we good with any CQE mod even if it doesn't align with #iterations? e.g. CQE mod == 100 and #iterations == 1024. The last CQE that we get is on the 1000th WR. What happens with the last 24?

dimasique1: > What happens with the last 24?

I need to investigate it, to be 100% sure. But I guess in current perftest we just don't get a CQ in such case and get out of the loop by another condition.

firasj: I thought about generalizing the CQ moderation but I think it will have a performance penalty.
Let's take for example post_list of size 4, and CQE every 3 WQEs. In such case, I can't do anything during init time (`ctx_set_send_reg_wqes()`) and in run time I will have to keep turning on and off the SIGNAL bit over all the post_list.

Also regarding my question:

> What happens with the last 24?

Current code always requests CQE on last iteration.

What do you suggest?

src/perftest_resources.c
firasj: Please note that `i` is the QP index and the post_lists are separate per QP.

src/perftest_parameters.c
dimasique1: 1) It is not bad that perftest can mimic an unoptimized case. We just need to make sure that it is not the default one :)
2) regarding the last question: looks like for this reason we have a condition on line 3823 to generate a CQe on the last iteration.

src/perftest_parameters.c
```diff
@@ -961,9 +961,15 @@ static void force_dependecies(struct perftest_parameters *user_param)
 	}
 
 	if (user_param->post_list > 1) {
-		user_param->cq_mod = user_param->post_list;
-		printf(RESULT_LINE);
-		printf("Post List requested - CQ moderation will be the size of the post list\n");
+		if (user_param->cq_mod == DEF_CQ_MOD) {
```
dimasique1: May be it worths checking that CQ_mod was explicitly set to DEF_CQ_MOD, but not chosen by default with a flag or smth.

src/perftest_parameters.c
firasj: Adding code to handle the generic case affects the default one: whenever we will have `post_list > 1`, we will have to check the signal bit on all WRs.

dimasique1: Not sure it can hurt the performance dramatically, since it is done only if the condition is hit.
Anyway, thank you for a great work, I will test it soon and merge if it works well!

firasj:
Any updates?

dimasique1:
> Any updates?

I see BW peak[MB/sec] zero now on both of the sides with default parameters on all bw benchmarks. Sorry that it takes so much time to give feedback.

firasj:
Hi @dimasique1 ,

Sorry for the delay. I've reuploaded the patch with a couple changes:
1. Fixed the loop condition.
2. Handled the case of explicit CQE (where `#tot_iterations % cq_mod != 0`).
3. Fixed a bug in the original code that fills tcompleted array.

dimasique1:
Hi @firasj,

Thank you for the update! I will be able to test the change this week. Will update you correspondingly.

firasj:
Hi,

Any updates? Is there anyway that I could help speed things up?

dimasique1:
Hi @firasj. I've checked the PR functionally, it seems to be working well, thanks for a great change. Sorry it took that much time. 

The only complain is a compilation warning on FreeBsd:

In file included from src/perftest_resources.c:29:
src/perftest_resources.h:118:9: warning: 'ALIGN' macro redefined [-Wmacro-redefined]
#define ALIGN(value, alignment) (((value) % (alignment) == 0) ?  \
        ^
/usr/include/machine/param.h:81:9: note: previous definition is here
#define ALIGN(p)                _ALIGN(p)

firasj:
> In file included from src/perftest_resources.c:29:
> src/perftest_resources.h:118:9: warning: 'ALIGN' macro redefined [-Wmacro-redefined]
> #define ALIGN(value, alignment) (((value) % (alignment) == 0) ?
> ^
> /usr/include/machine/param.h:81:9: note: previous definition is here
> #define ALIGN(p) _ALIGN(p)
> 
> Please, fix it.

Renamed macro to `ROUND_UP`.

> I also would be grateful, if the change you have added to the ctx_set_send_reg_wqes() can also be applied to ctx_set_send_exp_wqes(), otherwise it looks like CQ moderation is always equal to post_list size (if it is not 1) for experimental verbs.

I have applied the same change to experimental verbs, but I'm unable to test this. Can you please confirm it works?

> I plan to add a couple of commits regarding the CQ moderation soon, please, take them into consideration.

I'll keep an eye on it. Do you plan to merge them before this PR?

dimasique1:
@firasj thanks for a great effort.

Please, note that --post_list is not working if compiled with ./configure --enable-ibv_wr_api, but it is because it requires additional coding. So I will add a commit forbidding this flag so far with this api.

firasj:
Thank you @dimasique1.
One of my teammates is working on post_list with `--enable-ibv_wr_api` flag, and a PR will be published soon.

dimasique1:
@firasj Great, thanks! I hope I will be able to merge it faster than yours:)

PR #76, opened 2020-02-12
Support post list mode for receive WQEs

firasj:
Similar to TX post list, generate a post list of receive WQEs that are
chained and submitted together.

In order for this work, avoid posting receive WRs immediately after
completion and instead check for available space on QP.

Signed-off-by: Firas Jahjah <firasj@amazon.com>

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -254,8 +254,10 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 	}
 
 	if (tst == BW || tst == LAT_BY_BW) {
-		printf("  -l, --post_list=<list size>");
-		printf(" Post list of WQEs of <list size> size (instead of single post)\n");
+		printf("  -l, --post_list=<list size>\n");
+		printf(" Post list of send WQEs of <list size> size (instead of single post)\n");
+		printf("      --recv_post_list=<list size>");
+		printf(" Post list of receive WQEs of <list size> size (instead of single post)\n");
 	}
 
 	if (tst != FS_RATE) {
@@ -641,6 +643,7 @@ static void init_perftest_params(struct perftest_parameters *user_param)
 	user_param->iters		= (user_param->tst == BW && user_param->verb == WRITE) ? DEF_ITERS_WB : DEF_ITERS;
 	user_param->dualport		= OFF;
 	user_param->post_list		= 1;
+	user_param->recv_post_list	= 1;
 	user_param->use_srq		= OFF;
 	user_param->use_xrc		= OFF;
 	user_param->use_rss		= OFF;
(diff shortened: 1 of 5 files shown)


firasj:
Hi @dimasique1 ,
Sorry for the late response.
Yes, on EFA, we can see an improvement when posting RX WQEs in batches.
I'll close the PR for now until I fix the comments.

firasj:
Hello,

It's been a while but I've fixed your comments and rebased on master branch.

Thanks

dimasique1:
Hello @firasj,

Thanks for the contribution and fixing my previous comments. 

Seems like the flag works not as expected:
"./ib_send_bw -d mlx5_2 --recv_post_list=10" gets stuck. I guess it happens if (rx-depth % recv_post_list) != 0:

./ib_send_bw -d mlx5_2 --recv_post_list=10 -r 100 works well.

Please, fix or, if it is not possible, change rx-depth accordingly with a warning (if my guessing is correct).

Thanks!

firasj:
Hey @dimasique1,
Thanks for the quick response.
I have found and fixed the issue.

firasj:
Kind reminder.

src/perftest_parameters.c
```diff
@@ -2838,6 +2847,8 @@ void ctx_print_test_info(struct perftest_parameters *user_param)
 
 	if (user_param->post_list > 1)
 		printf(" Post List       : %d\n",user_param->post_list);
+	if (user_param->recv_post_list > 1)
+		printf(" Receive Post List : %d\n", user_param->recv_post_list);
```
dimasique1, line 2851: please change to " Recv Post List  : %d\n"

dimasique1:
@firasj,

I see the same issue with "./ib_send_bw -d mlx5_0 --recv_post_list=3", it just gets stuck. -r 30 doesn't help.

firasj:
I believe this is related to the known issue mentioned in README:
> 8. for postlist, tests will not finish correctly when using iterations and num_iters/poslist_val has a remainder.

Do you want me fail on initialization stage and let the user know?

dimasique1:
> I believe this is related to the known issue mentioned in README:
> 
> > 1. for postlist, tests will not finish correctly when using iterations and num_iters/poslist_val has a remainder.
> 
> Do you want me fail on initialization stage and let the user know?

Yeah, you are right. I would appreciate if you add such a check (if you can add it at the same place for send post_list, it would be just great) and fix the printing in case of the flag used (I mentioned in review). Everything else is just fine.

src/perftest_parameters.c
firasj: Done

firasj:
> > I believe this is related to the known issue mentioned in README:
> > > 1. for postlist, tests will not finish correctly when using iterations and num_iters/poslist_val has a remainder.
> > 
> > 
> > Do you want me fail on initialization stage and let the user know?
> 
> Yeah, you are right. I would appreciate if you add such a check (if you can add it at the same place for send post_list, it would be just great) and fix the printing in case of the flag used (I mentioned in review). Everything else is just fine.

Added a check for `post_list` in a separate commit, and handled the same check for `recv_post_list` in 2nd commit.

dimasique1:
@firasj 

Thank you for a wonderful contribution!

Issue #61, opened 2019-10-01
--use_cuda silently ignored when building without GPU support


drossetti:
This is bad, as people are convinced to be running a build with GPU support, so for example they report unreasonable performance.

HassanKhadour:
Hi drossetti, I see that the fix was merged.
Thank you

PR #74, opened 2020-01-30
SRD support for send WR API

dkkranz:
Introduce post list mode to IBV send WR API, and add support for SRD, including RDMA read.

PR #79, opened 2020-02-25
Removed EXP features and added DC support over rdma-core

dimasique1:


dimasique1:
Also new post send API become default

src/write_lat.c
```diff
@@ -80,6 +80,12 @@ int main(int argc, char *argv[])
 		return FAILURE;
 	}
 
+	/* In case of ib_write_bw, PCI relaxed ordering should be disabled since we're polling for data change
```
bureddy: s/ib_write_bw/ib_write_lat?

src/perftest_parameters.c
```diff
@@ -471,8 +471,8 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 		printf(" Set the Traffic Class in GRH (if GRH is in use)\n");
 
 		#ifdef HAVE_CUDA
-		printf("      --use_cuda ");
-		printf(" Use CUDA lib for GPU-Direct testing.\n");
+		printf("      --use_cuda=<cuda device id>");
```
drossetti: this part of the patch changes the syntax of the --use_cuda option while is not documented in the commit description.

src/perftest_parameters.c
```diff
@@ -667,7 +670,10 @@ static void init_perftest_params(struct perftest_parameters *user_param)
 	user_param->rate_limit_type	= DISABLE_RATE_LIMIT;
 	user_param->is_rate_limit_type  = 0;
 	user_param->output		= -1;
+#ifdef HAVE_CUDA
```
drossetti: this fragment and the others below fix #61 which should be documented

src/write_lat.c
```diff
@@ -80,6 +80,12 @@ int main(int argc, char *argv[])
...
+	 * of last packet so in case of relaxed odering we might get the last packet in wrong order thus the test
+	 * would be incorrect
+	 */
+	user_param.disable_pcir = 1;
```
drossetti: It is still interesting to benchmark write latency with RO enabled, which would be possible if you added a way to use write with immediate and ibv_poll_cq() on the receiver side.
Would you accept such a request?

src/perftest_parameters.c
```diff
@@ -471,8 +471,8 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 		printf(" Set the Traffic Class in GRH (if GRH is in use)\n");
 
 		#ifdef HAVE_CUDA
-		printf("      --use_cuda ");
-		printf(" Use CUDA lib for GPU-Direct testing.\n");
+		printf("      --use_cuda=<cuda device id>");
+		printf(" Use CUDA secific device for GPU-Direct testing\n");
```
drossetti: secific=specific
GPU-Direct = GPUDirect RDMA

src/perftest_parameters.c
spotluri: also, it will help to make it an optional argument and pick device zero if none is specified. There are quite a few places where this test is used in automation. Requiring an argument will force changes everywhere.

galpress:
> Also new post send API become default

Only mlx4, mlx5 and efa providers support that API.. Others are going to break?

src/perftest_resources.c
```diff
@@ -338,8 +311,14 @@ static inline int _new_post_send(struct pingpong_context *ctx,
 	default:
 		fprintf(stderr, "Post send failed: unknown operation code.\n");;
 	}
-
-	if (qpt == IBV_QPT_UD) {
+	if (qpt == IBV_QPT_DRIVER) {
```
galpress: IBV_QPT_DRIVER doesn't necessarily imply DC, SRD for ex. uses it as well (the code already exists, maybe these patches aren't rebased).
This function should test for connection type instead.

src/perftest_resources.c
```diff
@@ -2228,47 +1642,98 @@ struct ibv_qp* ctx_qp_create(struct pingpong_context *ctx,
...
+				fprintf(stderr, "Couldn't create rdma old QP - %s\n", strerror(errno));
+			}
+			else
+			{
+				qp = ctx->cm_id->qp;
+			}
+
 	} else if (user_param->connection_type == SRD) {
 		#ifdef HAVE_SRD
 		qp = efadv_create_driver_qp(ctx->pd, &attr, EFADV_QP_DRIVER_TYPE_SRD);
```
galpress: This is not rebased, but an if (use_old_post_send) should be added here as well to decide between efadv_create_driver_qp and efadv_create_qp_ex

src/perftest_resources.c
```diff
@@ -3804,11 +3810,15 @@ int run_iter_lat_send(struct pingpong_context *ctx,struct perftest_parameters *u
 		ctx->wr[0].sg_list->length = user_param->size;
 		ctx->wr[0].send_flags = 0;
 		#ifdef HAVE_IBV_WR_API
-		ctx->qpx[0]->wr_flags |= IBV_SEND_SIGNALED;
+		ctx->qpx[0]->wr_flags = 0;
```
galpress: Why was that moved?

src/write_lat.c
dimasique1: Yes, such a PR will be gladly accepted

src/write_lat.c
dimasique1: fixed, thanks

src/perftest_parameters.c
dimasique1: Fixed

src/perftest_parameters.c
dimasique1: We will try to figure out the best approach here, thanks for your feedback

src/perftest_resources.c
dimasique1: This will be covered by the backport of the https://github.com/linux-rdma/perftest/pull/74

src/perftest_resources.c
dimasique1: this will be covered by the backport of https://github.com/linux-rdma/perftest/pull/74

src/perftest_resources.c
dimasique1: thanks, looks like now it is fixed

dimasique1:
> > Also new post send API become default
> 
> Only mlx4, mlx5 and efa providers support that API.. Others are going to break?

Even mlx4 doesn't support, so it falls back to old one. The tool can also be recompiled with the corresponding flag, if needed.

PR #81, opened 2020-03-24
Fix CUDA runs with mr_per_qp flag

galpress:
When using CUDA with the mr_per_qp flag, buffer[0] is always used
instead of per QP buffer. Fix that by splitting the GPU init flow and
buffer allocation flow.

Signed-off-by: Gal Pressman <galpress@amazon.com>

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -63,10 +63,8 @@ struct check_alive_data check_alive_data;
 static CUdevice cuDevice;
 static CUcontext cuContext;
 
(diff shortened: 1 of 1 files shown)


galpress:
Kind reminder

dimasique1:
Hello Gal,
LGTM, I will merge when I test it properly, waiting for hardware so far. Sorry for long response.

dimasique1:
Thanks for the contribution, Gal!

PR #86, opened 2020-05-09
Falling back to old post_send API for devices not supporting ibv_wr_* one

dimasique1:
Only devices supporting ibv_wr_* API are in the whitelist to use it. Others fall back to old one without issues on QP creation.

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1394,12 +1394,25 @@ int verify_params_with_device_context(struct ibv_context *context,
 		}
 	}
 
-	if (current_dev == CONNECTX3_PRO || current_dev == CONNECTX3)
+	// those are devices supporting new post send
+	if (current_dev != CONNECTIB &&
+		current_dev != CONNECTX4 &&
+		current_dev != CONNECTX4LX &&
+		current_dev != CONNECTX5 &&
+		current_dev != CONNECTX5EX &&
+		current_dev != CONNECTX6 &&
+		current_dev != CONNECTX6DX &&
+		current_dev != CONNECTX6LX &&
+		current_dev != CONNECTX7 &&
+		current_dev != MLX5GENVF &&
+		current_dev != BLUEFIELD &&
+		current_dev != BLUEFIELD2 &&
+		current_dev != EFA)
 	{
 		if (!user_param->use_old_post_send)
 		{
-			fprintf(stderr, " Warning: ConnectX-3 and ConnectX-3 Pro don't support WR postsend API!\n");
-			fprintf(stderr, " Warning: Falling back to ibv_post_send() API\n");
+			fprintf(stderr, " Warning: The device chosen doesn't support WR post send API!\n");
+			fprintf(stderr, " Warning: Falling back to using ibv_post_send()\n");
 			user_param->use_old_post_send = 1;
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/perftest_resources.c
```diff
@@ -1394,12 +1394,25 @@ int verify_params_with_device_context(struct ibv_context *context,
 		}
 	}
 
-	if (current_dev == CONNECTX3_PRO || current_dev == CONNECTX3)
+	// those are devices supporting new post send
```
rleon, line 1397: The correct way to perform fall-back is to do test in configure stage, set some config define after that and use that define in the code.

Definitely not to whiltelist devices which won't allow to use this perftest for future devices.

Thanks

galpress: I'm not sure this can be resolved in configure time, we can check that rdma-core supports the new API, but how do you check if the provider supports it?

rleon: In theory, we can write simple application test that will try to use this API and if it gets EOPNOTSUP/ENOSYS, it will mean to us that it is not supported.

If it is not possible to do during configure stage, we can do this test inside perftest too, for first call, and cache the result.

dimasique1: I will work on testing inside the perftest, but I also want to state that whitelist is not such a big problem: anyway we have such a whitelist[ here](https://github.com/linux-rdma/perftest/blob/2490003cc8ecc33db053de0fefb023ea191dbec4/src/perftest_parameters.c#L1539).

rleon: > I will work on testing inside the perftest, but I also want to state that whitelist is not such a big problem: anyway we have such a whitelist[ here](https://github.com/linux-rdma/perftest/blob/2490003cc8ecc33db053de0fefb023ea191dbec4/src/perftest_parameters.c#L1539). And this whitelist is very important too, because we cannot do some stuff for unknown devices (such as inline).

One of the important goals of the library and various capabilities is to avoid whitelists.

dimasique1: > > I will work on testing inside the perftest, but I also want to state that whitelist is not such a big problem: anyway we have such a whitelist[ here](https://github.com/linux-rdma/perftest/blob/2490003cc8ecc33db053de0fefb023ea191dbec4/src/perftest_parameters.c#L1539). And this whitelist is very important too, because we cannot do some stuff for unknown devices (such as inline).
> 
> One of the important goals of the library and various capabilities is to avoid whitelists.

I understand. Do we have a cap for ibv_wr_* api support?

rleon: > > > I will work on testing inside the perftest, but I also want to state that whitelist is not such a big problem: anyway we have such a whitelist[ here](https://github.com/linux-rdma/perftest/blob/2490003cc8ecc33db053de0fefb023ea191dbec4/src/perftest_parameters.c#L1539). And this whitelist is very important too, because we cannot do some stuff for unknown devices (such as inline).
> > 
> > 
> > One of the important goals of the library and various capabilities is to avoid whitelists.
> 
> I understand. Do we have a cap for ibv_wr_* api support?

To some extend - yes.
Try to create QP with IBV_QP_INIT_ATTR_SEND_OPS_FLAGS flag, if you get an error, I think that it should be EOPNOTSUPP, you will know that it is not supported.

PR #88, opened 2020-05-10
Implemented proper fall-back to ibv_post_send without whitelists

dimasique1:


src/perftest_resources.c
```diff
@@ -1579,6 +1556,9 @@ int ctx_init(struct pingpong_context *ctx, struct perftest_parameters *user_para
 	for (i=0; i < user_param->num_of_qps; i++) {
 		if (create_qp_main(ctx, user_param, i, num_of_qps)) {
 			fprintf(stderr, "Failed to create QP.\n");
+			if (user_param->connection_type == DC) {
+				fprintf(stderr, "DC is not supported for the divice chosen.\n");
```
galpress, line 1560: Is this related to the patch? Also there's a typo in device.

src/perftest_resources.c
```diff
@@ -1618,9 +1598,9 @@ int create_reg_qp_main(struct pingpong_context *ctx,
 	}
 
 	if (ctx->qp[i] == NULL) {
-		fprintf(stderr, "Unable to create QP.\n");
-		return FAILURE;
-	}
+			return FAILURE;
```
galpress, line 1601: Same question, is this related?

dimasique1: > Same question, is this related?

Yes, it is, because we don't want to print this in fall-back case

src/perftest_resources.c
dimasique1: > Is this related to the patch? Also there's a typo in device.

Yes, it is, because the reason of failure of DC QP creation is fact that only new API supports DC.

HassanKhadour:

PR #91, opened 2020-06-18
Fix iWARP connection failures - Modify backlog passed to rdma_listen.

mkalderon:
The size of the backlog affects the time it takes to run rdma_listen in iWARP.
The iWARP stack allocates work entries based on the number.
When passing 0, the default is used which is 1K, this translates to the iWARP
rdma_listen taking ~40 MS leading to connection establishment timeouts
(when the client attempts to connect before the server is ready ).
There is no need to send the maximum backlog as the perftest is written in a
one pending connection establishment anyway, therefore, modify the rdma_listen
to use "1" as the backlog.

Signed-off-by: Michal Kalderon <michal.kalderon@marvell.com>

### DIFF
--- a/src/perftest_communication.c
+++ b/src/perftest_communication.c
@@ -1083,7 +1083,7 @@ int rdma_server_connect(struct pingpong_context *ctx,
 		return 1;
 	}
 
-	if (rdma_listen(ctx->cm_id_control,0)) {
+	if (rdma_listen(ctx->cm_id_control,1)) {
 		fprintf(stderr, "rdma_listen failed\n");
 		return 1;
 	}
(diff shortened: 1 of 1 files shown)


mkalderon:
> I have mixed feelings about this change.
> From one side, it looks like this change won't be a problem for RoCE/IB because they don't really care about backlog.
> From another side, the real fix is in kernel, kernel should change alloc_work_entries() to be based on memory pools, it will fix this problem.

Thanks Leon for the review. 
I think the two don't contradict. I can submit a kernel patch in addition, but the application is expected to work well on older kernels / distros as well. 
The fix suggested here is not wrong, functionality wise, very low risk, and is backward compatible to any older kernel and as you mentioned, won't affect RoCE / IB.

rleon:
No problem,

@dimasique1, I'm fine with this change, just run perftest over any non-iWARP network to double check and feel free to merge it.

Thanks

mkalderon:
@dimasique1 , is there anything else needed to merge this request?
thanks.

dimasique1:
@mkalderon Thanks for your contribution!

PR #94, opened 2020-06-21
Use IBV WR API for warm up iterations

dkkranz:
If perftest was configured with IBV WR API, use it for warmup iterations
instead of the post send verb.

Signed-off-by: Daniel Kranzdorf <dkkranzd@amazon.com>


dimasique1:
Thanks for the contribution!

PR #95, opened 2020-06-22
configure: fix ibv_wr_api enable argument

firasj:
The syntax for AC_ARG_ENABLE is:
  AC_ARG_ENABLE(option-name, help-string, action-if-present, action-if-not-present)

The mistake here is to consider the two actions as action-if-enabled and
action-if-disabled.

Scenario:
  ./configure --enable-ibv_wr_api
would end up running action-if-present and cause the new API to be
disabled.

Fix:
Since the default for autoconf when no action is defined is to set a
special variable named with the enable_ prefix, remove the custom
actions and let autoconf handle it.

Fixes: 705ed2e ("Made ibv_wr_api default during the build")
Signed-off-by: Firas Jahjah <firasj@amazon.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -43,11 +43,10 @@ AC_ARG_VAR(CUDA_H_PATH, help-string)
 AC_ARG_VAR(RANLIB, ranlib tool)
 
 AC_ARG_ENABLE([ibv_wr_api],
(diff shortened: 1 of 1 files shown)


dimasique1:
Thanks for the contribition!

PR #102, opened 2020-08-13
Resolve GID conflict with latest rdma-core upstream

SimonRaviv:


rleon:
NAK, it has no conflicts with real upstream version, but with some internal variant.

PR #106, opened 2020-09-27
Fix memory leak caused by MR copy.

biopuppet:
Just remove 2 lines of useless code.


dimasique1 (changes requested):
Hello @biopuppet 

Please, fix the commit message ("Update perftest_resources.c" is not representative).

Thanks!

PR #107, opened 2020-09-29
Set send function pointers in every latency test iteration

galpress:
This insures that inline check is done per iteration msg size (when
running with "-a"), and not once per all iterations.
The fix was done for the tests that set IBV_SEND_INLINE in old-api code.

Signed-off-by: Shadi Ammouri <sammouri@amazon.com>
Signed-off-by: Gal Pressman <galpress@amazon.com>

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -3857,6 +3857,13 @@ int run_iter_lat_write(struct pingpong_context *ctx,struct perftest_parameters *
 	}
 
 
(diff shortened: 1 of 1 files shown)


dimasique1 (commented):
Don't we have same code at lines 3849 and 4053?

galpress:
Oops.. Yes we do, you fixed it in f9b3b87d3ec7 ("Fixed latency at low message size with --all flag").
Ignore this PR 😄 .

PR #109, opened 2020-10-04
Fix typo in perftest_communication error output

dariusgrassi:
Hi, I found a little typo when running some RDMA benchmarks with the perftest utilities (ib_read_bw and ib_write_bw). Seemed like it wasn't intentional.


dimasique1:
Thanks for the contribution!

PR #112, opened 2020-10-22
Add additional Marvell (QLogic) devices to support inline feature

mkalderon:
Additional devices should be added to the list of devices supporting inline

Signed-off-by: Michal Kalderon <michal.kalderon@marvell.com>


mkalderon:
Hi @dimasique1 , 
Could please merge this pull request, it's only a few additional devices supporting inline, 
riskless :-) 

thanks,
Michal

Issue #116, opened 2020-12-31
Should disable PCIe Relaxed Ordering by default?


Honggang-LI:
If perftest was built with a machine which PCIe Relaxed Ordering compliant, and run over two machines which are **_NOT_** PCIe Relaxed Ordering compliant, we will see **significantly performance degradation** for large size message.

https://bugzilla.redhat.com/show_bug.cgi?id=1902855

The perftest of "ib_send_lat RC" test on data size of 4194304 bytes and 8388608 bytes had test time increase of about **250 fold** when compared with the same test on other MLX4 ROCE device, like MLX5 CX-3.

So, should we disable PCIe RO by default, and let the user to enable it or not when run the perftest benchmarks?

Thanks

rleon:
I think that the best way will be to check PCIe RO during perftest execution and disable/enable accordingly.

Thanks

Honggang-LI:
> I think that the best way will be to check PCIe RO during perftest execution and disable/enable accordingly.

Unfortunately, I don't know which C function or library can be used to check PCIe RO. Do you have a suggestion?

In the _"perftest/configure.ac"_ file, PCIe RO will be disabled if specific vendor and device ID was found.
```bash
if [test $IS_FREEBSD = no]; then
        CPU_IS_RO_COMPLIANT=yes
        # Actually this is check for being affected by a known issue
        # with Intel CPUs:
        # https://lore.kernel.org/patchwork/patch/820922/
        intel_no_ro_devices="6f01 6f02 6f03 6f04 6f05 6f06 6f07 6f08 6f09 6f0a 6f0b 6f0c 6f0d 6f0e 2f01 2f02 2f03 2f04 2f05 2f06 2f07 2f08 2f09 2f0a 2f0b 2f0c 2f0d 2f0e"
        for cpu_num in $intel_no_ro_devices; do
                ro_nonsupported=$(lspci -nn | grep -i 8086:$cpu_num)
                if [ test "$ro_nonsupported" ]; then
                        CPU_IS_RO_COMPLIANT=no
                fi
        done
fi

AC_TRY_LINK([
#include <infiniband/verbs.h>],
        [int x = IBV_ACCESS_RELAXED_ORDERING;],[HAVE_RO=yes], [HAVE_RO=no])
AM_CONDITIONAL([HAVE_RO],[test "x$HAVE_RO" = "xyes"])
[...]

PR #117, opened 2021-01-11
Check PCIe relaxed ordering compliant

Honggang-LI:
Significantly performance degradation may be observed when PICe
relaxed ordering enabled over CPU which is not PCIe RP compliant.
Emit a warning message for such scenario.

https://github.com/linux-rdma/perftest/issues/116

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -174,6 +174,8 @@ if [test $HAVE_SNIFFER = yes]; then
 fi
 
 if [test $IS_FREEBSD = no]; then
+	AC_CHECK_HEADERS([pci/pci.h],,[AC_MSG_ERROR([pciutils header files not found, consider installing pciutils-devel])])
+	AC_CHECK_LIB([pci], [pci_init], [LIBPCI=-lpci], AC_MSG_ERROR([libpci not found]))
 	CPU_IS_RO_COMPLIANT=yes
 	# Actually this is check for being affected by a known issue
 	# with Intel CPUs:
@@ -193,6 +195,7 @@ AC_TRY_LINK([
 AM_CONDITIONAL([HAVE_RO],[test "x$HAVE_RO" = "xyes"])
 if [test $HAVE_RO = yes] && [test "x$CPU_IS_RO_COMPLIANT" = "xyes"]; then
         AC_DEFINE([HAVE_RO], [1], [Enable Relaxed Ordering])
+	LIBS=$LIBS" -lpci"
 fi
 
 AC_TRY_LINK([#include <infiniband/verbs.h>],
--- a/perftest.spec
+++ b/perftest.spec
@@ -8,6 +8,7 @@ Source:         http://www.openfabrics.org/downloads/%{name}-%{version}.tar.gz
 Url:            http://www.openfabrics.org
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  libibverbs-devel librdmacm-devel libibumad-devel
+BuildRequires:  pciutils-devel
 
 %description
 gen3 uverbs microbenchmarks
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -11,6 +11,10 @@
 #include "perftest_parameters.h"
 #include "raw_ethernet_resources.h"
 #include<math.h>
+#ifdef HAVE_RO
+#include <stdbool.h>
+#include <pci/pci.h>
+#endif
 #define MAC_LEN (17)
 #define ETHERTYPE_LEN (6)
 #define MAC_ARR_LEN (6)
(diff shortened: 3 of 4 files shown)


rleon:
@dimasique1 i didn't review deeply this PR, but the idea is right one. +1

Thanks

src/perftest_parameters.c
```diff
@@ -2873,6 +2907,11 @@ void ctx_print_test_info(struct perftest_parameters *user_param)
 	printf(" Connection type : %s\t\tUsing SRQ      : %s\n", connStr[user_param->connection_type], user_param->use_srq ? "ON"  : "OFF");
 	#ifdef HAVE_RO
 	printf(" PCIe relax order: %s\n", user_param->disable_pcir ? "OFF"  : "ON");
+	if ((check_pcie_relaxed_ordering_compliant() == false) &&
+	    (user_param->disable_pcir == 0)) {
+		printf(" WARNING: CPU is not PCIe relaxed ordering compliant.\n");
+		printf(" WARNING: You should disable PCIe RO for both server and clinet.\n");
```
dimasique1, line 2913: Please, add to the warning the flag name --disable_pcie_relaxed, so user doesn't need to look through the help

configure.ac
```diff
@@ -60,10 +60,12 @@ AC_PROG_LIBTOOL
 AC_PROG_RANLIB
 AC_HEADER_STDC
 AC_CHECK_HEADERS([infiniband/verbs.h],,[AC_MSG_ERROR([ibverbs header files not found])])
+AC_CHECK_HEADERS([pci/pci.h],,[AC_MSG_ERROR([pciutils header files not found])])
```
dimasique1, line 63: Please, add check here and below that it is not a FreeBSD machine as we don't support relaxed ordering there. Let's also update README that it is a new dependency on Linux

perftest.spec
```diff
@@ -8,6 +8,7 @@ Source:         http://www.openfabrics.org/downloads/%{name}-%{version}.tar.gz
 Url:            http://www.openfabrics.org
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  libibverbs-devel librdmacm-devel libibumad-devel
+BuildRequires:  pciutils-devel
```
dimasique1, line 11: this doesn't influence FreeBSD, right?

dimasique1 (changes requested):

dimasique1:
Thanks for the contribution, please, see my comments.

src/perftest_parameters.c
Honggang-LI: Yes, I will do it.

configure.ac
Honggang-LI: Will move _AC_CHECK_HEADERS_ and _AC_CHECK_LIB_ into this section:
```
 if [test $IS_FREEBSD = no]; then
+       AC_CHECK_HEADERS([pci/pci.h],,[AC_MSG_ERROR([pciutils header files not found])])
+       AC_CHECK_LIB([pci], [pci_init], [LIBPCI=-lpci], AC_MSG_ERROR([libpci not found]))
        CPU_IS_RO_COMPLIANT=yes
        # Actually this is check for being affected by a known issue
        # with Intel CPUs:
```

perftest.spec
Honggang-LI: No,  as FreeBSD application management does not rpm based.

Honggang-LI:
Please review new patch. Thanks

src/perftest_parameters.c
```diff
@@ -163,6 +167,36 @@ static int get_cache_line_size()
...
+#ifdef HAVE_RO
+/******************************************************************************
+  Check PCIe Relaxed Ordering
+
+  Stolen from https://github.com/pciutils/pciutils/blob/master/example.c
+ ******************************************************************************/
+static bool check_pcie_relaxed_ordering_compliant(void) {
+	struct pci_access *pacc;
+	struct pci_dev *dev;
+	unsigned int c;
```
dimasique1, line 179: warning: variable 'c' set but not used,

Please, remove

configure.ac
```diff
@@ -174,6 +174,8 @@ if [test $HAVE_SNIFFER = yes]; then
 fi
 
 if [test $IS_FREEBSD = no]; then
+	AC_CHECK_HEADERS([pci/pci.h],,[AC_MSG_ERROR([pciutils header files not found])])
+	AC_CHECK_LIB([pci], [pci_init], [LIBPCI=-lpci], AC_MSG_ERROR([libpci not found]))
```
dimasique1, line 178: please add 
LIBS=$LIBS" -lpci"

as currently it doesn't compile with "./autogen.sh ; ./configure ; make -j"

dimasique1 (changes requested):
Thanks! Found a couple of more issues during the testing.

src/perftest_parameters.c
Honggang-LI: Fixed. thanks

configure.ac
Honggang-LI: fixed. thanks

configure.ac
```diff
@@ -174,6 +174,8 @@ if [test $HAVE_SNIFFER = yes]; then
 fi
 
 if [test $IS_FREEBSD = no]; then
+	AC_CHECK_HEADERS([pci/pci.h],,[AC_MSG_ERROR([pciutils header files not found])])
```
dimasique1, line 177: please, let it be like that:
"pciutils header files not found, consider installing pciutils-devel"

src/perftest_parameters.c
```diff
@@ -2873,6 +2905,11 @@ void ctx_print_test_info(struct perftest_parameters *user_param)
 	printf(" Connection type : %s\t\tUsing SRQ      : %s\n", connStr[user_param->connection_type], user_param->use_srq ? "ON"  : "OFF");
 	#ifdef HAVE_RO
 	printf(" PCIe relax order: %s\n", user_param->disable_pcir ? "OFF"  : "ON");
+	if ((check_pcie_relaxed_ordering_compliant() == false) &&
+	    (user_param->disable_pcir == 0)) {
+		printf(" WARNING: CPU is not PCIe relaxed ordering compliant.\n");
+		printf(" WARNING: You should disable PCIe RO with `--disable_pcie_relaxed` for both server and clinet.\n");
```
dimasique1, line 2911: cliENt

dimasique1 (changes requested):
Please, fix the wording

dimasique1:
@Honggang-LI 
Thank you for the contribution

PR #119, opened 2021-01-14
Print max msg size change to STDOUT

dkkranz:
When running RUN_ALL method, max message size change is expected, thus
should not be printed to stderr.

Signed-off-by: Daniel Kranzdorf <dkkranzd@amazon.com>


dimasique1:
Hello @dkkranz,

I agree with that, but it probably can influence users' automation, so I suggest to keep it as is.

HassanKhadour:
Hi @ddkranz,

I prefer to print the "max message size" to stderr, because it's still kind of error.

PR #122, opened 2021-05-28
use config.{sub,guess} from system

yetist:
The `config/config.{sub,guess}` files are too old and have not been updated after running `./autogen.sh`.

The easy way is to add the --force option to automake: `automake --foreign --add-missing --copy --force`.

But by doing so, I can't see the effect of the original old files, so I suggest to delete them directly.


talatb:
Thank you for your change, we will test it on aarch64 system and other archs and update.

HassanKhadour:
Hi, Thank you, We tested it and everything worked correctly,It's ready to be merged.

yetist:
Can this PR be merged?

rleon:
> @HassanKhadour
> 
> Can this PR be merged? Thanks.

It looks like it failed through cracks.

PR #124, opened 2021-08-24
Do not check if CPU is RO compliant at compile time

nmorey:
Patch d157d4246abe (Check PCIe relaxed ordering compliant) introduced run-time check
for relaxed ordering compliance. Thus there shouldn't be any compile time checks.
Current code is at risk on behaving differently on the same machine depending on which
CPU it was packaged.

Signed-off-by: Nicolas Morey-Chaisemartin <nmoreychaisemartin@suse.com>


nmorey:
@Honggang-LI Just to keep you in the loop

PR #127, opened 2021-10-11
SRD CQ notifications

gal-pressman:
Add support for SRD CQ notifications.
The first patch fixes a bug where the same completion channel is used for both TX and RX and messes up the test.

### DIFF
--- a/src/raw_ethernet_resources.c
+++ b/src/raw_ethernet_resources.c
@@ -1005,8 +1005,7 @@ int run_iter_fw(struct pingpong_context *ctx,struct perftest_parameters *user_pa
 		}
 
 		if (user_param->use_event) {
-
-			if (ctx_notify_events(ctx->channel)) {
+			if (ctx_notify_send_recv_events(ctx)) {
 				fprintf(stderr, "Failed to notify events to CQ");
 				return_value = FAILURE;
 				goto cleaning;
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -1406,11 +1406,6 @@ static void force_dependecies(struct perftest_parameters *user_param)
 			fprintf(stderr, " SRD does not support RDMA_CM\n");
 			exit(1);
 		}
-		if (user_param->use_event == ON) {
-			printf(RESULT_LINE);
-			fprintf(stderr, " SRD does not support events\n");
-			exit(1);
-		}
 		user_param->cq_mod = 1;
 	}
 
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1073,13 +1073,9 @@ int destroy_ctx(struct pingpong_context *ctx,
 		test_result = 1;
 	}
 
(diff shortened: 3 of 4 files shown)


amitgeron:
Hi,

Can you please add more details what issue the first patch actually resolves?

Thanks

galpress:
IIRC, the test uses the same completion channel and does:

- Send packet
- Remote side receives the packet and send another packet
- Receive packet from remote

And assumes that the next event is necessarily from the send completion, but it could theoretically be from the receive completion.

@firasj might remember the exact details.

amitgeron:
The unidirectional BW test for the client (run_iter_bw) is not supposed to receive data at all, so an event should never be delivered for the recv_cq, and the server side (run_iter_bw_server) isn't supposed to send any data, so an event should never be delivered for the send_cq.

Did you encounter any bug or issues when running the tests?

galpress:
It was a real issue, some tests (send_lat for example) do both send and receive, and they might fail without this fix.

mrgolin:
@HassanKhadour Can you please merge this PR or suggest some extra improvements?
Thanks!

sshaulnv:
@firasj could you please provide the case that will reproduce this issue with the command you used?

HassanKhadour:

galpress:
@firasj can you please provide the command lines we were running?
@sshaulnv what is needed to get this merged? Do you want me to test this on mlx5 hardware?

HassanKhadour:
Hi Gal, yes we need the commands in order to test it on mlx5 hardware

firasj:
@gal-pressman Can you please close this PR for now?
It's been on a low priority and we'll reopen once we have a reproduction and an answer.

mrgolin:
Hi, running:
```
for i in {1..20}; do ib_send_lat -c SRD -x 0 -F --report_gbits -a -n 10000 -r 1 -e & ib_send_lat -c SRD -x 0 -F --report_gbits -a -n 10000 -r 1 -e localhost; done
```
is reproducing the issue on EFA HW.
@sshaulnv @gal-pressman can you please test this with mlx5 HW?

src/raw_ethernet_resources.c
```diff
@@ -1005,8 +1005,7 @@ int run_iter_fw(struct pingpong_context *ctx,struct perftest_parameters *user_pa
 		}
 
 		if (user_param->use_event) {
-
-			if (ctx_notify_events(ctx->channel)) {
+			if (ctx_notify_send_recv_events(ctx)) {
```
sshaulnv, line 1008: @gal-pressman, can you describe this change please?

mrgolin:
@HassanKhadour Can you please merge this?

HassanKhadour:
Hi @mrgolin , conflicts should be resolved to merge.

mrgolin:
@HassanKhadour I've rebased and pushed as new PR https://github.com/linux-rdma/perftest/pull/211.

PR #92, opened 2020-06-19
Add ipv6 address support for initial communication.

dmonakhov:
Currenty initial negotiation performed via ipv4 which is not suitable for modern ipv6 only topology
This patch allow to specify which address family to use, default behaviour not changed.

New option:  --ipv6-addr

Usage example:
# Server
./ib_write_bw -d mlx5_0 --ipv6-addr
# Client
./ib_write_bw -d mlx5_0 --ipv6-addr 2a02:6b8:c0e:97f:0:441d:9fbd:3f1e

PR #129, opened 2021-12-20
Update PR #92 (add ipv6 support)

souravzzz:
Update PR #92 to fix issues

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,149 @@
+perftest (4.4-0.29-8-gba04f65) unstable; urgency=low
+
+  * Add SRD support for new IBV send WR API
+  * Forbid messages larger than SRD max msg size
+  * Add SRD support for RDMA read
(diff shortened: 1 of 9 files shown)


tbarbette:
Hey! Thanks for the v6 support :)
I tried the following but get the same message than mainline : 


```
sudo ./ib_write_bw --ipv6 -d mlx5_0 -x 3 -F --report_gbits -p 1800 -s 64 -q 1 -D 10 "babe:2::5"
Address family for hostname not supported for babe:2::5:1800
Problem in resolving basic address and port
Unable to open file descriptor for socket connection Unable to init the socket connection
```
What arguments should I add? I also tried giving the IPv6 address with [] but did not work better.
`ucmatose -s babe:2::5` works for instance.

HassanKhadour:

HassanKhadour:
Closing, functionality was added in https://github.com/linux-rdma/perftest/pull/218

Issue #49, opened 2019-02-01
Fails to build on mips*/armel/armhf: get_cycles not implemented for this architecture


bdrung:
perftest fails to build from source on mips*/armel/armhf (on Debian):
```
[...]
make  all-am
make[2]: Entering directory '/«BUILDDIR»/perftest-3.4+0.6.gc3435c2'
gcc -DHAVE_CONFIG_H -I.   -Wdate-time -D_FORTIFY_SOURCE=2  -g -Wall -D_GNU_SOURCE -O3 -g -O2 -fdebug-prefix-map=/«BUILDDIR»/perftest-3.4+0.6.gc3435c2=. -fstack-protector-strong -Wformat -Werror=format-security -c -o src/get_clock.o src/get_clock.c
In file included from /usr/include/arm-linux-gnueabihf/sys/time.h:21:0,
                 from src/get_clock.c:43:
/usr/include/features.h:148:3: warning: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE" [-Wcpp]
 # warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"
   ^~~~~~~
In file included from src/get_clock.c:48:0:
src/get_clock.h:101:2: warning: #warning get_cycles not implemented for this architecture: attempt asm/timex.h [-Wcpp]
 #warning get_cycles not implemented for this architecture: attempt asm/timex.h
  ^~~~~~~
src/get_clock.h:102:23: fatal error: asm/timex.h: No such file or directory
 #include <asm/timex.h>
                       ^
compilation terminated.
Makefile:786: recipe for target 'src/get_clock.o' failed
make[2]: *** [src/get_clock.o] Error 1
```

Debian bug: https://bugs.debian.org/863158

HassanKhadour:
Solved with merging https://github.com/linux-rdma/perftest/pull/134

PR #134, opened 2022-01-25
support get_cycles for riscv

kuroro99:
This patch support the get_cycles for riscv to get the system time. And then
perftest can be built successfully in riscv architecture, which was tested in
openeuler-riscv64 and debian-riscv64-qemu.


kuroro99:
This would resolve this issue https://github.com/linux-rdma/perftest/issues/49#issue-405619071, when built in riscv.

PR #152, opened 2022-06-30
configure: Cleanup CUDA configury

rajachan:
The current approach to enabling CUDA support in the bandwidth
benchmarks with an environment variable that takes that path to an
include file is extremely fragile.  As a low-hanging fruit, this commit
brings the CUDA configuration closer to ROCm, but relying on a flag to
enable the feature and AC_SEARCH_LIBS to update lib flags, or on
--with-cuda allowing a user to provide a custom installation path.
Ideally, we should rely on dlopen()'ing the symbols we need from these
libraries so a single build of perftest can work on systems with and
without CUDA SDK, but that's to come later.

Signed-off-by: Raghu Raja <raghu@enfabrica.net>


rajachan:
It is hard to tell who is maintaining perftests these days, but tagging @HassanKhadour who made the last stable release for review. I do not have triage permissions on this repo to tag reviewers.

sshaulnv:
By removing the include of the 'CUDA_PATH' in perftest_parameters.h (line 70) the commit is breaking the backward compatibility.

rajachan:
Indeed, I will fix that up. I am in favor of completely removing it if we can.. given the explicit configure warning (which we can turn into an error) and perhaps a call out in the changelog when this makes it into a release.

HassanKhadour:

galpress:
Does perftest.spec need a change as well?

sshaulnv:
we would move to loading cuda with dlopen by the end of the 25.07 release.

PR #154, opened 2022-07-05
Perftest: fix ending conditions for non-standard -t/-q/-n values

bsobczak:
It was revealed that in certain combination of tx_depth, connection
number, cq_mod and iteration values the code wrongly determines
ending conditions.

Especially, in the while loop condition a wrong comparison was used
in which a mix of signed and unsigned variables are summed or
subtracked without minding the resulting sign:

(ctx->scnt[index] - ctx->ccnt[index] + user_param->post_list) <=
    (user_param->tx_depth)
In this case when scnt is lower than ccnt by more than post_list
we end up with a high number compared to tx_depth just because
the scnt is unsigned.

In another case, for multiple qp involved, when cq_mod is not a divider
of iters we may end up in situation when:
  totccnt >= tot_iters
and there is an i-th connection with ctx->ccnt[$i] < iters
In such scenario the test will not be able to complete.

Commands for reproduction:
ib_read_bw -R -s 1024 -n 10000 -q 5 -t 12
ib_send_bw -a -b -R -t 83 -r 41
OS: FreeBSD 12/13

Sponsored-by: Intel Corporation
Signed-off-by: Bartosz Sobczak <bartosz.sobczak@intel.com>

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -3326,7 +3326,7 @@ int run_iter_bw(struct pingpong_context *ctx,struct perftest_parameters *user_pa
 				burst_iter = 0;
 			}
 			while ((ctx->scnt[index] < user_param->iters || user_param->test_type == DURATION) &&
-					(ctx->scnt[index] - ctx->ccnt[index] + user_param->post_list) <= (user_param->tx_depth) &&
+					(ctx->scnt[index] + user_param->post_list) <= (user_param->tx_depth + ctx->ccnt[index]) &&
 					!((user_param->rate_limit_type == SW_RATE_LIMIT ) && is_sending_burst == 0)) {
 
 				if (ctx->send_rcredit) {
@@ -3416,8 +3416,15 @@ int run_iter_bw(struct pingpong_context *ctx,struct perftest_parameters *user_pa
 							goto cleaning;
 						}
 
-						ctx->ccnt[wc_id] += user_param->cq_mod;
-						totccnt += user_param->cq_mod;
+						if (ctx->ccnt[wc_id] + user_param->cq_mod <= user_param->iters) {
+							ctx->ccnt[wc_id] += user_param->cq_mod;
+							totccnt += user_param->cq_mod;
+						} else {
+							int fill = user_param->iters - ctx->ccnt[wc_id];
+
+							ctx->ccnt[wc_id] += fill;
+							totccnt += fill;
+						}
 						if (user_param->noPeak == OFF) {
 							if (totccnt > tot_iters)
 								user_param->tcompleted[user_param->iters*num_of_qps - 1] = get_cycles();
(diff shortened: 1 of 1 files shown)


firasj:
> In this case when scnt is lower than ccnt

How can you get more completions than WQEs you have posted?

> when cq_mod is not a divider of iters

I thought we fail if the number of iterations is not a multiple of `post_list`, and if `post_list` is not a multiple of `cq_mod` and therefore iterations must be a multiple of `cq_mod`. Can you please clarify?

bsobczak:
> > nor the input parameters are validated?
> 
> See https://github.com/linux-rdma/perftest/blob/master/src/perftest_parameters.c#L1121-L1137. Although we might be missing the same check for `user_param->recv_post_list`.

thanks, missed that, however, my example of failure shows 'post_list=1', so that is escaping these checks. and so my example of commands in commit description also tend to operate on tx_depth, but post_list is standrad (1)

src/perftest_resources.c
```diff
@@ -3416,8 +3416,15 @@ int run_iter_bw(struct pingpong_context *ctx,struct perftest_parameters *user_pa
 							goto cleaning;
 						}
 
-						ctx->ccnt[wc_id] += user_param->cq_mod;
-						totccnt += user_param->cq_mod;
+						if (ctx->ccnt[wc_id] + user_param->cq_mod <= user_param->iters) {
```
HassanKhadour, line 3419: Couldn't this if else condition cause performance issues / degradations ?

HassanKhadour:
Hi @bsobczak Thanks for your contribution!
I understand the issues you are encountering and agree with you that it must be fixed.
But I'm worried about the if else added in run_iter_bw because it may cause performance issues since its in middle
of the performance measuring

src/perftest_resources.c
bsobczak: i didn't notice, but you are right, someone could see a little overhead.
Do you have any suggestions?
i think i see a few options to make it a little better:

1. add starting condition, that cq_mod must be a divider of iters
2. do something like:
```
int fill = min(cq_mod, user_param->iters - ctx->ccnt[wc_id]);
ctx->ccnt[wc_id] += fill;
totccnt += fill;
```
but i suppose the if statement would be just hidden here...
3. instead of doing this for every iteration in this loop, do the check after the loop and alter it if needed. i'd add something like this into line 3442:
```
for(j=0;j<num_of_qps;j++) {
  if (ctx->ccnt[j] > user_param->iters) {
    totccnt -= user_param->iters - ctx->ccnt[j];
    ctx->ccnt[j] =user_param->iters;
  }
}

``` 
even though there's an additional loop, some time should be saved for several reasons: in noPeak==OFF scenario it would be done after completion time is gathered;
in situations when more than 1 completion is polled we'd get from 2 to tx_depth times less checking.

alternatively, we could avoid the for loop and put it in line 3319, but i'm not sure we wouldn't still prematurely escape the while loop at line 3314

Any other thoughts?

sshaulnv:
Hi @bsobczak,
When setting a duration greater then 1 the results are reset.
command for example: ib_read_bw --duration 5

It's related in some way to the 'duration_param->state' that change to START_STATE during the run while it should be SAMPLE_STATE.

HassanKhadour:

HassanKhadour:
PR was merged with fixes

PR #156, opened 2022-08-01
perftest_parameters: Enable RDMA atomic with ODP feature

daimatsuda:
RDMA atomic with ODP is already enabled in the mlx5 driver.
There is no reason to block the feature.
cf. https://github.com/torvalds/linux/commit/17d2f88f92ce39b348f125f6b2e6eeb6b0906ac7

Signed-off-by: Daisuke Matsuda <matsuda-daisuke@fujitsu.com>

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1569,17 +1569,23 @@ int destroy_ctx(struct pingpong_context *ctx,
 static int check_odp_support(struct pingpong_context *ctx)
 {
 	struct ibv_device_attr_ex dattr;
-	int odp_support_send = IBV_ODP_SUPPORT_SEND;
-	int odp_support_recv = IBV_ODP_SUPPORT_RECV;
 	int ret = ibv_query_device_ex(ctx->context, NULL, &dattr);
 
 	if (ret) {
 		fprintf(stderr, " Couldn't query device for on-demand paging capabilities.\n");
 		return 0;
-	} else if (!(dattr.odp_caps.per_transport_caps.rc_odp_caps & odp_support_send)) {
+	}
+
+	/* These capabilities must be set by device drivers. */
+	if (!(dattr.odp_caps.general_caps & IBV_ODP_SUPPORT)) {
+		fprintf(stderr, " The device does not support On-Demand Paging.\n");
+                return 0;
+        }
+
+	if (!(dattr.odp_caps.per_transport_caps.rc_odp_caps & IBV_ODP_SUPPORT_SEND)) {
 		fprintf(stderr, " Send is not supported for RC transport.\n");
 		return 0;
-	} else if (!(dattr.odp_caps.per_transport_caps.rc_odp_caps & odp_support_recv)) {
+	} else if (!(dattr.odp_caps.per_transport_caps.rc_odp_caps & IBV_ODP_SUPPORT_RECV)) {
 		fprintf(stderr, " Receive is not supported for RC transport.\n");
 		return 0;
 	}
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1566,29 +1566,102 @@ int destroy_ctx(struct pingpong_context *ctx,
  *
  ******************************************************************************/
 #ifdef HAVE_EX_ODP
-static int check_odp_support(struct pingpong_context *ctx)
+static int check_odp_transport_caps(struct perftest_parameters *user_param, uint32_t caps)
+{
+	static char conn_str[][7] = {"RC", "UC", "UD", "RawEth", "XRC", "DC", "SRD"};
+	int conn = user_param->connection_type;
+	VerbType verb = user_param->verb;
+	MachineType machine = user_param->machine;
+	int duplex = user_param->duplex;
+
+	if (verb == SEND && duplex == OFF && machine == CLIENT) {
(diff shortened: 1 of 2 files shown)


daimatsuda:
Hello @HassanKhadour, thank you for taking a look.

> but I see that removing the if condition that blocks the odp with atomic will result in
> a wrong fail message for mlx4 devices (for example cx3pro)

Nice catch! I did some tests using the rxe driver, which does not have the ODP feature either,
and found that the wrong message is generated not only with atomic but also with other
operations when odp flag is specified.

> Need to find a way to enable/disable this feature according to the device support of odp

I have added another patch to handle this issue.

src/perftest_resources.c
```diff
@@ -1566,23 +1566,41 @@ int destroy_ctx(struct pingpong_context *ctx,
...
-		fprintf(stderr, " Receive is not supported for RC transport.\n");
+	}
+
+	/* These capabilities must be set by device drivers. */
+	if (!(dattr.odp_caps.general_caps & IBV_ODP_SUPPORT)) {
+		fprintf(stderr, " The device does not support On-Demand Paging.\n");
 		return 0;
 	}
+
+	if (verb == SEND && (
```
HassanKhadour, line 1585: Thanks ! ,Great way of checking!, but we miss here reading the capabilities for other transports (DC, UD ,etc..), can you please add them?

src/perftest_resources.c
```diff
@@ -1659,7 +1715,7 @@ int create_single_mr(struct pingpong_context *ctx, struct perftest_parameters *u
 	/* ODP */
 	#ifdef HAVE_EX_ODP
 	if (user_param->use_odp) {
-		if ( !check_odp_support(ctx) )
+		if ( !check_odp_support(ctx, user_param->verb, user_param->connection_type) )
```
HassanKhadour, line 1718: Thanks , looks great.
can you please pass only the user_param and use the pointers in check_odp_support?

daimatsuda: Thanks for the review!
I've done it.

daimatsuda:
Well, I found the way to fetch xrc odp caps and that there are some points to be improved.
I will repost the third patch today.

daimatsuda:
I think now the patches are OK. However, there seems a problem with the mlx5 driver, firmware, or hardware. I am using "Mellanox Technologies MT28908 Family [ConnectX-6]". The driver fills the capabilities at internal_fill_odp_caps().
cf. https://github.com/torvalds/linux/blame/master/drivers/infiniband/hw/mlx5/odp.c#L321

The problem is that some capabilities are not filled while the operation can be used with ODP. For example, I can use RC ATOMIC with ODP and can observe ODP code is working, but IB_ODP_SUPPORT_ATOMIC(0x10) is not set in the capability(rc_odp_caps). 
Here are the capabilities printed at the bottom of internal_fill_odp_caps():

> rc_odp_caps=2f / uc_odp_caps=0 / ud_odp_caps=1 / xrc_odp_caps=2d

It seems RC ATOMIC, UD SRQ_RECV, XRC RECV, XRC ATOMIC are left unfilled.
I doubt if MLX5_CAP_ODP() macro is working correctly with all capabilities, or the firmware may not configure the values properly. I have no time nor ideas to investigate this further right now. Do you have any idea about this problem?

Anyway, I believe checking ODP capabilities is necessary procedure. Merging the patches would not bother anybody because ODP ATOMICs are not available in perftest from the beginning. If the problem is reproducible on other environments, then it should be addressed by the vendor.

src/perftest_resources.c
```diff
@@ -1566,29 +1566,129 @@ int destroy_ctx(struct pingpong_context *ctx,
...
 		return 0;
-	} else if (!(dattr.odp_caps.per_transport_caps.rc_odp_caps & IBV_ODP_SUPPORT_RECV)) {
-		fprintf(stderr, " Receive is not supported for RC transport.\n");
+
+	case UD:
+		if (verb == SEND && machine == CLIENT) {
+			if (!(dattr.odp_caps.per_transport_caps.ud_odp_caps & IBV_ODP_SUPPORT_SEND)) {
+				fprintf(stderr, " ODP Send is not supported for UD transport.\n");
+				return 0;
+			}
```
talatb, line 1625: We see a lot of duplication code in this change, can we have it in better shape ? 
like create a func that can do this check and print generated string.

daimatsuda: Thanks for the comment! I modified the patch.

MalekImam (commented):
`
if (verb == SEND && machine == CLIENT) {
		if (!(caps & IBV_ODP_SUPPORT_SEND)) {
			fprintf(stderr, " ODP Send is not supported for %s transport.\n", conn_str[conn]);
			return 0;
		}
	} else if (verb == SEND && machine == SERVER) {
		if (!(caps &
		      (IBV_ODP_SUPPORT_RECV | IBV_ODP_SUPPORT_SRQ_RECV))) {
			fprintf(stderr, " ODP Recv is not supported for %s transport.\n", conn_str[conn]);
			return 0;
		}
	}
`

Great job, but I see that the check for the bidirectional case is missing when using the send verb in line 1585.

HassanKhadour:

daimatsuda:
I have added another block to check the bidirectional case.
Thanks for your suggestion!

HassanKhadour:

PR #176, opened 2022-11-08
Perftest: Add new post send and inline support for HNS

HelloTcc:
HNS roce has been accepted by the rdma-core for a few years now, but it was not supported by Perftest. For this reason, the inline and the new post send of hns cannot be tested in perftest, even though they have been merged to the rdma-core mainline for some time.

In this PR，the first patch add hns roce to the device list. And make it support the new post send and set a default inline size.
The second patch increase the max_inline_data to 1024 to support test for some hns roce device.

### DIFF
--- a/src/perftest_parameters.h
+++ b/src/perftest_parameters.h
@@ -178,8 +178,8 @@
 #define UC_MAX_RX     (16000)
 #define MIN_CQ_MOD    (1)
 #define MAX_CQ_MOD    (1024)
-#define MAX_INLINE    (912)
-#define MAX_INLINE_UD (884)
+#define MAX_INLINE    (1024)
+#define MAX_INLINE_UD (1024)
 #define MIN_EQ_NUM    (0)
 #define MAX_EQ_NUM    (2048)
 
--- a/src/perftest_parameters.h
+++ b/src/perftest_parameters.h
@@ -381,6 +381,7 @@ enum ctx_device {
 	QLOGIC_AHP		= 27,
 	BLUEFIELD3		= 28,
 	ERDMA			= 29,
+	HNS			= 30,
 };
 
 /* Units for rate limiter */
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -1861,6 +1861,13 @@ enum ctx_device ib_dev_name(struct ibv_context *context)
 			case 61344 : dev_fname = EFA; break;
 			case 61345 : dev_fname = EFA; break;
 			case 4223  : dev_fname = ERDMA; break;
+			case 41506 : dev_fname = HNS; break;
+			case 41507 : dev_fname = HNS; break;
+			case 41508 : dev_fname = HNS; break;
+			case 41509 : dev_fname = HNS; break;
+			case 41510 : dev_fname = HNS; break;
+			case 41512 : dev_fname = HNS; break;
+			case 41519 : dev_fname = HNS; break;
 			default	   : dev_fname = UNKNOWN;
 		}
 	}
@@ -2066,6 +2073,8 @@ static void ctx_set_max_inline(struct ibv_context *context,struct perftest_param
 				user_param->inline_size = 128;
 			else if (current_dev == ERDMA)
 				user_param->inline_size = 96;
+			else if (current_dev == HNS)
+				user_param->inline_size = 32;
 
 		} else {
 			user_param->inline_size = 0;
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -2042,7 +2042,8 @@ int verify_params_with_device_context(struct ibv_context *context,
 		current_dev != BLUEFIELD &&
 		current_dev != BLUEFIELD2 &&
 		current_dev != BLUEFIELD3 &&
-		current_dev != EFA)
+		current_dev != EFA &&
+		current_dev != HNS)
 	{
 		if (!user_param->use_old_post_send)
 		{


src/perftest_parameters.h
```diff
@@ -178,8 +178,8 @@
 #define UC_MAX_RX     (16000)
 #define MIN_CQ_MOD    (1)
 #define MAX_CQ_MOD    (1024)
-#define MAX_INLINE    (912)
-#define MAX_INLINE_UD (884)
+#define MAX_INLINE    (1024)
+#define MAX_INLINE_UD (1024)
```
sshaulnv, line 182: This change can cause problems with devices with smaller max inline size.
the ibv_create_qp() will return NULL value if the MAX_INLINE isn't supported.

HelloTcc: At present, these two macros are only used for parameter verification to ensure that the inline length specified by the user through "- I" does not exceed this value.  By default, such a large inline size is not directly used. By the way, the current value are not supported for some HWs, but it works well.

In the current rdma core, if the driver does not support the value set by the user, the user will be informed of the max inline size that the driver can support in the returned attr.cap.max_inline_data. There is a example：https://github.com/linux-rdma/perftest/pull/178/commits/99decb7a2d1b1b12f985b04a9fa64f1ae25dfae3

HelloTcc: > This change can cause problems with devices with smaller max inline size. the ibv_create_qp() will return NULL value if the MAX_INLINE isn't supported.

Hi @sshaulnv 
As described in the link, the ibv_create_qp_ex returns the actual QP value via qp_init_attr_ex.
https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/man/ibv_create_qp_ex.3

So, the value returned by the driver (in the qp_init_attr_ex.cap) is more suitable for determining whether the packet can be sent in inline mode.

src/perftest_parameters.h
```diff
@@ -178,8 +178,8 @@
 #define UC_MAX_RX     (16000)
 #define MIN_CQ_MOD    (1)
 #define MAX_CQ_MOD    (1024)
-#define MAX_INLINE    (912)
-#define MAX_INLINE_UD (884)
+#define MAX_INLINE    (1024)
```
HassanKhadour, line 181: I saw your response to sshaulnv and I debugged it, When exceeding  those values the qp creation in ctx_qp_create method is failing and returning null with errno 28 and the max_inline_data in the attr is equal to the inline_size specified by the user so its not even entering the if condition that checks the inline_size.
Can you please send your setup information?

HelloTcc: I test it in a 5.10 kernel and rdma-core v41 version, it works well for our hns roce. 

And I tested the ConnectX-4 Lx in the same environment, and the interesting thing is that it still works well beyond the original limit. Until the test to the length of 1005, it has the same problem as you describe

The normal log:
```bash
[root@localhost perftest]# ib_send_lat -d mlx5_0 -I 1004 127.0.0.1 &
[2] 1766247
[root@localhost perftest]# ---------------------------------------------------------------------------------------
                    Send Latency Test
 Dual-port       : OFF          Device         : mlx5_0
 Number of qps   : 1            Transport type : IB
 Connection type : RC           Using SRQ      : OFF
 PCIe relax order: ON
                    Send Latency Test
 Dual-port       : OFF          Device         : mlx5_0
 Number of qps   : 1            Transport type : IB
 Connection type : RC           Using SRQ      : OFF
 PCIe relax order: ON
 ibv_wr* API     : ON
 TX depth        : 1
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 3
 Max inline data : 1004[B]
 rdma_cm QPs     : OFF
 Data ex. method : Ethernet
 ibv_wr* API     : ON
 RX depth        : 512
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 3
 Max inline data : 1004[B]
 rdma_cm QPs     : OFF
 Data ex. method : Ethernet
 local address: LID 0000 QPN 0x0133 PSN 0x9a60a6
 local address: LID 0000 QPN 0x0132 PSN 0x432e72
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:192:168:100:100
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:192:168:100:100
 remote address: LID 0000 QPN 0x0132 PSN 0x432e72
 remote address: LID 0000 QPN 0x0133 PSN 0x9a60a6
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:192:168:100:100
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:192:168:100:100
 #bytes #iterations    t_min[usec]    t_max[usec]  t_typical[usec]    t_avg[usec]    t_stdev[usec]   99% percentile[usec]   99.9% percentile[usec]
 #bytes #iterations    t_min[usec]    t_max[usec]  t_typical[usec]    t_avg[usec]    t_stdev[usec]   99% percentile[usec]   99.9% percentile[usec]
 2       1000          1.08           3.19         1.20                1.20             0.06            1.38                    3.19
 2       1000          1.08           3.91         1.19                1.20             0.06            1.40                    3.91

[1]-  Done                    ib_send_lat -d mlx5_0 -I 1004
[2]+  Done                    ib_send_lat -d mlx5_0 -I 1004 127.0.0.1
```
The exception log：
```bash
[root@localhost perftest]# ib_send_lat -d mlx5_0 -I 1005 &
[1] 1766854
[root@localhost perftest]#
************************************
* Waiting for client to connect... *
************************************

[root@localhost perftest]# ib_send_lat -d mlx5_0 -I 1005 127.0.0.1 &
[2] 1766865
[root@localhost perftest]# Unable to create QP.
Failed to create QP.
 Couldn't create IB resources
Unable to create QP.
Failed to create QP.
 Couldn't create IB resources

[1]-  Exit 1                  ib_send_lat -d mlx5_0 -I 1005
[2]+  Exit 1                  ib_send_lat -d mlx5_0 -I 1005 127.0.0.1
```

I guess this problem should not be directly related to my mr. This may be caused by some limitations of mlx5. I briefly looked at the code of mlx5, and I feel that the most likely possibility is that the ENOSPC error is reported in the `alloc_var_entry()`.

HelloTcc: > I guess this problem should not be directly related to my pr. This may be caused by some limitations of mlx5. I briefly looked at the code of mlx5, and I feel that the most likely possibility is that the ENOSPC error is reported in the `alloc_var_entry()`.

The `alloc_var_entry()` seems to be for DV(Direct verbs) and I'm not familiar with that. So,  I use CM to test, so that the mlx5 driver enters the traditional create qp path. Like this:
```bash
ib_send_lat -d mlx5_0 -I 1024 -R &
ib_send_lat -d mlx5_0 -I 1024 192.168.100.100 -R &
```

Then I analyzed the problem and finally found that the problem in the CM scenario can be solved by the following patch (for rdma-core).
```patch
diff --git a/providers/mlx5/verbs.c b/providers/mlx5/verbs.c
index 0331ea9..0194d44 100644
--- a/providers/mlx5/verbs.c
+++ b/providers/mlx5/verbs.c
@@ -1597,6 +1597,8 @@ static int mlx5_calc_send_wqe(struct mlx5_context *ctx,
        if (attr->cap.max_inline_data) {
                inl_size = size + align(sizeof(struct mlx5_wqe_inl_data_seg) +
                        attr->cap.max_inline_data, 16);
+               if (inl_size > ctx->max_sq_desc_sz)
+                       inl_size = ctx->max_sq_desc_sz;
        }

        if (attr->comp_mask & IBV_QP_INIT_ATTR_MAX_TSO_HEADER) {
```

HelloTcc: Hi，@HassanKhadour，From the above phenomenon, it seems that the problem is not directly related to the change of max inline size. Is this patch acceptable?

HassanKhadour (approved):
Thanks for your contribution!

Issue #89, opened 2020-05-11
Initializing huge buffers with rand() takes too much time causing iWARP MPA timeout.


yuvalbason:
During MPA negotiation, when passive side gets MPA receive packet, the RDMA-CM stack
creates connection resources (PD, MR, CQ, QP) and initiates MPA offload.
The active side, who initiated the MPA connection sets a timer that times out in case of huge buffers
(initiating those buffers takes more than reasonable MPA timeout).
I think that in this case, perftest should support an option to skip buffer initialization -
	for (i = 0; i < ctx->buff_size; i++) {
				((char*)ctx->buf[qp_index])[i] = (char)rand();

dimasique1:
Yuval,
Thanks for the contribution!
Such a PR will be appreciated.

PR #177, opened 2022-11-08
Perftest: replace rand() with getrandom() during MR buffer initialization

HelloTcc:
rand() has very poor performance in some OS.

ib_send_bw will spend a lot of time during MR initialization when testing large packects in above scenario.

A simple test has been done:
```c
#define HUGE_MR_SIZE 2147483647
int main(int argc, char *argv[])
{
        char *a = malloc(HUGE_MR_SIZE * sizeof(char));
        unsigned int i;
        char *tmp = a;
        int ret;

        srand(time(NULL));
        if (a == NULL)
                exit(1);

        if (argc <= 1)
                goto fall_back;

        for (i = HUGE_MR_SIZE; i > 0;) {
                ret = getrandom(tmp, i, 0);
                if (ret < 0)
                        goto fall_back;
                tmp += ret;
                i -= ret;
        }
        goto out;

fall_back:
        for(i = 0; i < HUGE_MR_SIZE; i++)
                a[i] = (char)rand();
out:
        free(a);
        return 0;
}
```
```bash
time ./a.out
real    5m35.033s
user    5m33.546s
sys     0m0.918s

time ./a.out 1

real    0m6.454s
user    0m0.000s
sys     0m6.449s
```
As shown in the test above, getrandom() has a much better performance, so replace rand() with it. 
The test in perftest has also been done.

This patch fixes https://github.com/linux-rdma/perftest/issues/89

Signed-off-by: Chengchang Tang <tangchengchang@huawei.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -60,6 +60,7 @@ AC_PROG_LIBTOOL
 AC_PROG_RANLIB
 AC_HEADER_STDC
 AC_CHECK_HEADERS([infiniband/verbs.h],,[AC_MSG_ERROR([ibverbs header files not found])])
+AC_CHECK_HEADERS([sys/random.h],,)
 AC_CHECK_LIB([ibverbs], [ibv_get_device_list], [], [AC_MSG_ERROR([libibverbs not found])])
 AC_CHECK_LIB([rdmacm], [rdma_create_event_channel], [], AC_MSG_ERROR([librdmacm-devel not found]))
 AC_CHECK_LIB([ibumad], [umad_init], [LIBUMAD=-libumad], AC_MSG_ERROR([libibumad not found]))
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -22,6 +22,9 @@
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
+#ifdef HAVE_SYS_RANDOM_H
+#include <sys/random.h>
+#endif
 #ifdef HAVE_SRD
 #include <infiniband/efadv.h>
 #endif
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -60,6 +60,7 @@ AC_PROG_LIBTOOL
 AC_PROG_RANLIB
 AC_HEADER_STDC
 AC_CHECK_HEADERS([infiniband/verbs.h],,[AC_MSG_ERROR([ibverbs header files not found])])
+AC_CHECK_HEADERS([sys/random.h],,[AC_MSG_ERROR([getrandom available])])
```
HassanKhadour, line 63: "getrandom available" is uncorrect message, should be not available/found instead

HassanKhadour (changes requested):
Thanks for your contribution!

configure.ac
HelloTcc: Thanks，my mistake.

configure.ac
```diff
@@ -60,6 +60,7 @@ AC_PROG_LIBTOOL
 AC_PROG_RANLIB
 AC_HEADER_STDC
 AC_CHECK_HEADERS([infiniband/verbs.h],,[AC_MSG_ERROR([ibverbs header files not found])])
+AC_CHECK_HEADERS([sys/random.h],,[AC_MSG_ERROR([getrandom not found])])
```
HassanKhadour, line 63: Sorry for this, I just checked witha  setup that doesn't have sys/random.h and it failed to compile,I think it should be checked with ac_try_link because now if we don't have sys/random.h the configuration will fail and the HAVE_SYS_RANDOM_H will not help us in this situation

HassanKhadour (changes requested):

HassanKhadour (approved):

configure.ac
HassanKhadour: great, it solved the issue.
Thanks for your contribution.

PR #178, opened 2022-11-08
Perftest: Fix verification of max_inline_data for *_create_qp_ex()

HelloTcc:
As described in the link, the ibv_create_qp_ex returns the actual QP value via qp_init_attr_ex：
https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/man/ibv_create_qp_ex.3

Currently, qp_init_attr.cap.max_inline_data is used for validation both in *_create_qp() and *_create_qp_ex(). But actually, when entering the create_qp_ex path, the variable qp_init_attr is not used. So the current check for the *_create_qp_ex() branch is meaningless.

The qp_init_attr_ex.cap.max_inline_data is used to check the max_inline_data in *_create_qp_ex() path in this patch. And related printing error has also been fixed.

Fixes: 13f71777e6f0 ("Added new post_send API usage for RC,UC,UD,XRC")
Signed-off-by: Chengchang Tang <tangchengchang@huawei.com>

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -2437,22 +2437,21 @@ struct ibv_qp* ctx_qp_create(struct pingpong_context *ctx,
 	int dc_num_of_qps = user_param->num_of_qps / 2;
 
 	int is_dc_server_side = 0;
+	struct ibv_qp_init_attr attr;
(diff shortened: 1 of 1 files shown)


HassanKhadour:
Hi HelloTcc, thanks for your contribution!

PR #180, opened 2022-11-17
Support additional EFA device ids

mrgolin:
Add 0xefa2-0xefa4 device IDs to the devices database.

Signed-off-by: Chen Brasch <cbrasch@amazon.com>


gal-pressman:
Which devices are 0xefa3 and 0xefa4?
The kernel and rdma-core only support 0xefa0 to 0xefa2..

mrgolin:
This was to support upcoming devices, missed that there is a device dependent logic.

Removed 0xefa3 and 0xefa4.

PR #183, opened 2022-11-25
Fix tx_depth when test method is RUN_INFINITELY

DillionApple:
The original if statement will force the tx_depth to be less than iters, which is 1000 by default, even if the user runs with --run_infinitely option in which case the iters option is useless. So a user may not get what it wants when setting --run_infinitely and -t at the same time.

Signed-off-by: DillionApple <DillionMango@gmail.com>


HassanKhadour:
Thanks for your contribution!

PR #185, opened 2023-01-06
Perftest: remove unsupported feature option configuration

changchengx:
In 4948417abacf ("Removed some features that are not supported in rdma-core"), HAVE_ACCL_VERBS feature has been removed.


HassanKhadour:
Thanks for your contribution

PR #187, opened 2023-01-08
Perftest: correct comment for check_link & check_mtu

changchengx:
Signed-off-by: Liu, Changcheng <jerrliu@nvidia.com>


HassanKhadour:
Thanks for your contribution

PR #190, opened 2023-01-08
Perftest: correct logic to destroy mr for per qp

changchengx:
In ad2e127f4610, mr_index is used to audited the created MR. mr_index should be only audited when mr_per_qp feature is used. fix the logic to audit MR numbers.


HassanKhadour:
Thanks for your contribution

PR #191, opened 2023-01-09
Perftest: remove dead code and adjust comment place for readability

changchengx:


HassanKhadour:
Thanks for your contribution

PR #196, opened 2023-01-10
Perftest: enable inherit environment LDFLAGS

changchengx:
Sometimes, it needs to use environment to specify the library path to be linked during development stage if the library doesn't exist in system installed path. enable inherit environment LDFLAGS in configure to support it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -310,6 +310,7 @@ if [test $HAVE_MLX5DV_LIB = yes] && [test $HAVE_MLX5DV = yes]; then
 fi
 
 CFLAGS="-g -Wall -D_GNU_SOURCE -O3 $CFLAGS"
+LDFLAGS="$LDFLAGS"
 LIBS=$LIBS" -lpthread"
 AC_SUBST([LIBUMAD])
 AC_SUBST([LIBMATH])


HassanKhadour:
Hi, doesn't this PR override the LDFLAGS in line 237?

changchengx:
@HassanKhadour 
1. If there's no LDFLAGS environment is set when running ```configure``` script, this PR doesn't change anything

2. If there's LDFLAGS environment is set when running ```configure``` script by user.
    1) If ```--with_rocm``` option isn't used, [line 237](https://github.com/linux-rdma/perftest/pull/196/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R237) won't be called. This PR only set LDFLAGS to the user's LDFLAGS configuration.
    2) If ```--with_rocm``` option is set, [line 237](https://github.com/linux-rdma/perftest/pull/196/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R237) will be called. It means that user's LDFLAGS configuration is added at the end of LDFLAGS.

HassanKhadour:
Thanks,

So it means if we use with_rocm, LDFLAGS="$LDFLAGS" will take the value from line 237? and it will be LDFLAGS=-L$with_rocm/lib64 -Wl,-rpath=$with_rocm/lib64 -L$with_rocm/lib -Wl,-rpath=$with_rocm/lib -lamdhip64 $LDFLAGS
?

HassanKhadour:
Thanks for your contribution!

PR #198, opened 2023-02-16
Perftest: add null mr option

flx42:
On a single system, using a null MR can be used to generate unidirectional PCIe traffic to or from the device. Without a null mr, PCIe traffic is generated to and from the device and might saturate memory bandwidth.

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1925,6 +1925,14 @@ int create_single_mr(struct pingpong_context *ctx, struct perftest_parameters *u
 		}
 	}
 
+	if (user_param->use_null_mr) {
+		ctx->null_mr = ibv_alloc_null_mr(ctx->pd);
+		if (!ctx->null_mr) {
+			fprintf(stderr, "Couldn't create null MR\n");
+			return FAILURE;
+		}
+	}
+
 	if (ctx->is_contig_supported == SUCCESS)
 		ctx->buf[qp_index] = ctx->mr[qp_index]->addr;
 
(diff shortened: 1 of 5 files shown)


flx42:
One thing I wasn't sure about: `--use-null-mr` VS `--use_null_mr`, I couldn't see an obvious convention across the existing arguments.

Also, this patch was verified with a tool to measure PCIe traffic, I verified that there was only PCIe writes to the root complex with `ib_write_bw`, and only PCIe reads to the root complex with `ib_read_bw`.

sshaulnv:
Hi @flx42,
1. Can this feature also be used with SEND verb? if not, need to disable it and print a message. The current implementation may be a bit confusing.
2. need to add it to the man pages.

flx42:
Thanks @sshaulnv, I discussed with @bureddy and he suggested to enable the feature only for read and write verbs for the sake of simplicity. If that's a problem I can do the testing for SEND.

Anyway I should have addressed your two suggestions now, please take a look.

src/perftest_resources.c
```diff
@@ -3304,6 +3312,9 @@ void ctx_set_send_reg_wqes(struct pingpong_context *ctx,
 				(user_param->connection_type == RawEth) ? (user_param->size - HW_CRC_ADDITION) : user_param->size;
 
 			ctx->sge_list[i*user_param->post_list + j].lkey = ctx->mr[i]->lkey;
+			if (user_param->use_null_mr && (user_param->verb == WRITE || user_param->verb == READ)) {
```
sshaulnv, line 3315: after adding the restriction in the perftest_parameters.c, you can remove the check of the verb type from here.

flx42: Good catch, thanks, updated.

HassanKhadour:
Thanks, for contribution!

PR #204, opened 2023-03-17
perftest: Add Intel device names and inline data sizes

tatyana-en:
Add two Intel devices and their corresponding
inline data sizes.


HassanKhadour:
Hi @tatyana-en,
Thanks for the PR, please resolve the conflicts.

PR #205, opened 2023-04-20
extend --use-null-mr option for server side

bureddy:
specifying --user-null-mr option at the server side in READ/WRITE tests will perform remote memory read/writes from HCA.


bureddy:
@flx42 FYI

HassanKhadour:
Hi @bureddy 

Thanks for your contribution!, Please add more detailed explanation in the commit message

sshaulnv:
@bureddy, since now its not client support only, please adapt the help section and the man page (`"Allocate a null memory region for the client with ibv_alloc_null_mr."`).

PR #210, opened 2023-05-18
Modify --source_ip to --bind_sounce_ip to fix init connection establishment with specific interface

w180112:
When there are several network interface with different subnet address, perftest tools will always choose default route even I add --source_ip option to ask perftest to bind an interface.
I found there are two options use same name called "--source_ip". Therefore I add a prefix "raw_" for raw eth series options to avoid further using two same option name when adding new options.

BR,
Huai-En, Tseng

### DIFF
--- a/man/perftest.1
+++ b/man/perftest.1
@@ -268,7 +268,7 @@ many different options and modes.
  Use IPv6 GID. Default is IPv4.
  Not relevant for RawEth.
 .TP
-.B --source_ip
+.B --bind_source_ip
  Source IP of the interface used for connection establishment. By default taken from routing table.
  Not relevant for RawEth.
 .TP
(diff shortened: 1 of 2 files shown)


HassanKhadour:
Hi @w180112 Thanks for the contribution!
please fix also the man page

w180112:
Hi @HassanKhadour,
Thanks for feedback. I already add these into man page.
BR,

HassanKhadour:
Hi @w180112,
Thank you for your PR.

After checking with relevant people, We concluded that this change will break the backward compatibility for many platforms and users.
In order not to create damage for many people, I see that changing only the source ip is a better option in this case.

I apologize for any confusion or inconvenience caused, thanks for your understanding.

w180112:
Hi @HassanKhadour,

Thanks your feedback!
I modify to just modify the --source_ip option to --bind_source_ip.

BR,

HassanKhadour:
Thanks!, can you please just add this change to the commit message?

w180112:
Hi @HassanKhadour,

Thanks your comment, it is already done.

BR,

PR #211, opened 2023-05-21
SRD CQ notifications

mrgolin:
Add support for SRD CQ notifications.
The first patch fixes a bug where the same completion channel is used for both TX and RX and messes up the test.

Replaces PR https://github.com/linux-rdma/perftest/pull/127.


HassanKhadour:
Hi @mrgolin , Thanks for your contribution, merged!

PR #218, opened 2023-07-02
Add ipv6 address support for initial communication.

HassanKhadour:
Currenty initial negotiation performed via ipv4 which is not suitable for modern ipv6 only topology.
This patch allow to specify which address family to use, default behaviour not changed.

New option:  --ipv6-addr

Usage example:
./ib_write_bw -d mlx5_0 --ipv6-addr
./ib_write_bw -d mlx5_0 --ipv6-addr 2a02:6b8:c0e:97f:0:441d:9fbd:3f1e

### DIFF
--- a/src/perftest_communication.c
+++ b/src/perftest_communication.c
@@ -908,9 +908,8 @@ int set_up_connection(struct pingpong_context *ctx,
 int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters *user_param)
 {
 	char *service;
-	int temp,num_of_retry= NUM_OF_RETRIES;
-	struct sockaddr_storage source_sin;
-	struct sockaddr *sin;
+	int temp, num_of_retry = NUM_OF_RETRIES;
+	struct sockaddr_storage sin, source_sin;
 	struct sockaddr *source_ptr = NULL;
 	struct addrinfo *res;
 	struct rdma_cm_event *event;
@@ -931,8 +930,14 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters 
 		return FAILURE;
 	}
 
-	sin = res->ai_addr;
-	sockaddr_set_port(sin, (unsigned short)user_param->port);
+	if (res->ai_addr->sa_family == AF_INET) {
+		memcpy(&sin, res->ai_addr, sizeof(struct sockaddr_in));
+	}
+	else {
+		memcpy(&sin, res->ai_addr, sizeof(struct sockaddr_in6));
+	}
+	sockaddr_set_port((struct sockaddr *)&sin, (unsigned short)user_param->port);
+	freeaddrinfo(res);
 
 	if (user_param->has_source_ip) {
 		if (check_add_port(&service, 0x0, user_param->source_ip, &hints, &res))
@@ -942,8 +947,14 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters 
 		}
 		memset(&source_sin, 0x0, sizeof(source_sin));
 		//coverity[deref_after_free]
-		memcpy(&source_sin, res->ai_addr, sizeof(source_sin));
+		if (res->ai_addr->sa_family == AF_INET) {
+			memcpy(&source_sin, res->ai_addr, sizeof(struct sockaddr_in));
+		}
+		else {
+			memcpy(&source_sin, res->ai_addr, sizeof(struct sockaddr_in6));
+		}
 		source_ptr = (struct sockaddr *)&source_sin;
+		freeaddrinfo(res);
 	}
 	while (1)
 	{
@@ -953,14 +964,12 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters 
 			return FAILURE;
 		}
 
-		if (rdma_resolve_addr(ctx->cm_id, source_ptr, sin, 2000)) {
-			freeaddrinfo(res);
+		if (rdma_resolve_addr(ctx->cm_id, source_ptr, (struct sockaddr *)&sin, 2000)) {
 			fprintf(stderr, "rdma_resolve_addr failed\n");
 			return FAILURE;
 		}
 
 		if (rdma_get_cm_event(ctx->cm_channel,&event)) {
-			freeaddrinfo(res);
 			fprintf(stderr, "rdma_get_cm_events failed\n");
 			return FAILURE;
 		}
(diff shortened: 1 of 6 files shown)


dmonakhov:
LGTM

src/perftest_communication.c
```diff
@@ -906,29 +909,30 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters
 {
 	char *service;
 	int temp,num_of_retry= NUM_OF_RETRIES;
-	struct sockaddr_in sin, source_sin;
+	struct sockaddr_storage source_sin;
+	struct sockaddr *sin;
 	struct sockaddr *source_ptr = NULL;
 	struct addrinfo *res;
```
xiakun-microsoft, line 915: Now I get to know why the initial author kept a copy for both source and dest addresses respectively, `res` is reused and freed right after the address in it is copied out.
suggest:
```c
	struct sockaddr_storage sin, source_sin;
	struct sockaddr *sin_ptr = NULL, *source_ptr = NULL;
	struct addrinfo *res;
```

src/perftest_communication.c
```diff
@@ -1124,35 +1132,36 @@ int rdma_server_connect(struct pingpong_context *ctx,
 	struct rdma_conn_param conn_param;
 	struct addrinfo hints;
 	char *service;
-	struct sockaddr_in sin;
+	struct sockaddr *sin;
```
xiakun-microsoft, line 1135: keep a copy of the destination address, so `res` can be freed earlier, `freeaddrinfo(res)` needn't to be called more than once.
suggest changing it to `struct sockaddr_storage sin`, and casting it `struct sockaddr *` when used

xiakun-microsoft (changes requested):

src/perftest_communication.c
HassanKhadour: Sorry, I want to be sure that I understand you correctly, do you mean to keep the old implementation ? (memcpy(&sin, res->ai_addr, sizeof(sin))) and then free the res regularly right after in addition to changing sin to sockaddr_storage and casting when needed?

src/perftest_communication.c
HassanKhadour: can you please explain and elaborate more about this comment, specially about the sin_ptr and its usage?

src/perftest_communication.c
```diff
@@ -940,7 +944,6 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters
 		//coverity[deref_after_free]
 		memcpy(&source_sin, res->ai_addr, sizeof(source_sin));
```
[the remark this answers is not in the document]
HassanKhadour: ack

src/perftest_communication.c
```diff
@@ -906,29 +909,30 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters
...
 	}
-	memcpy(&sin, res->ai_addr, sizeof(sin));
-	freeaddrinfo(res);
-	sin.sin_port = htons((unsigned short)user_param->port);
+
+	sin = res->ai_addr;
+	sockaddr_set_port(sin, (unsigned short)user_param->port);
 
 	if (user_param->has_source_ip) {
 		if (check_add_port(&service, 0x0, user_param->source_ip, &hints, &res))
```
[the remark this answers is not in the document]
HassanKhadour: Thanks, I understand you, do you suggest to keep the old implementation for it "memcpy(&sin, res->ai_addr, sizeof(sin));" and then using res freely while casting sin when needed? if no, can you please tell what is your suggestion?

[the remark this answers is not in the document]
xiakun-microsoft: Yes, I suggest keeping the old implementation

src/perftest_communication.c
xiakun-microsoft: Looked at the code again, `sin_ptr` is unnecessary, you can safely cast `struct sockaddr_storage source_sin` to `struct sockaddr *`.  `struct sockaddr *source_ptr = NULL;` is needed to pass to `rdma_resolve_addr` when the source_ip is not supplied.

The target of these addresses is to get cm_id from [rdma_resolve_addr(3)](https://linux.die.net/man/3/rdma_resolve_addr). `rdma_resolve_addr` requires the source (optional) and dest addresses to be `struct sockaddr *`.
In the old code, `struct sockaddr_in sin, source_sin;` are used as the storage for the dest and source addresses respectively that are copied from `struct addrinfo *res`. 

revise my suggestion to:
```c
	struct sockaddr_storage sin, source_sin;
	struct sockaddr *source_ptr = NULL;
	struct addrinfo *res;
```

`sin` and `source_sin` are used to store the addresses from `struct addrinfo *res`. When `rdma_resolve_addr` is called, cast `sin` to `struct sockaddr *`, `source_ptr` is already a `struct sockaddr *` pointer to either NULL or `&source_sin`

src/perftest_communication.c
xiakun-microsoft: yes, keep the old implementation.
refer to my another comment:
https://github.com/linux-rdma/perftest/pull/218/files#r1254053570

HassanKhadour:
Thanks guys for the review.
@xiakun-microsoft Would appreciate if you can do again.

src/perftest_communication.c
```diff
@@ -938,7 +947,12 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters
 		}
 		memset(&source_sin, 0x0, sizeof(source_sin));
 		//coverity[deref_after_free]
-		memcpy(&source_sin, res->ai_addr, sizeof(source_sin));
+		if (((struct sockaddr *)&sin)->sa_family == AF_INET) {
```
xiakun-microsoft, line 950: this line will work, but `sin` is not related to code around here, better to use `res->ai_addr->sa_family == AF_INET`

src/perftest_communication.c
```diff
@@ -969,6 +983,7 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters
 
 		if (event->event != RDMA_CM_EVENT_ADDR_RESOLVED) {
 			fprintf(stderr, "unexpected CM event %d\n",event->event);
+			freeaddrinfo(res);
```
xiakun-microsoft, line 986: forgot to remove this one?

HassanKhadour: right

src/perftest_communication.c
HassanKhadour: fixed

HassanKhadour:
@xiakun-microsoft , thanks for the review, fixed, can you please check it again?

xiakun-microsoft (approved):
LGTM. Thanks for working on it.

HassanKhadour:
Thanks for your review and help guys, merged

HassanKhadour:
Hi,

We noticed a degradation in this PR in some cases.
We reverted this PR locally to try to point to the issue, after investigation we saw that adding extra dummy parameter into "static const struct option long_options" in perftest_parameters.c will cause a degradation in some cases, e.g. 
"ib_send_bw -p 19891 -d mlx5_0 -q 1 -s 1024 --report_gbits --output=bandwidth -R -c RC --duration 10"

just would like to pay your attention to this issue.

PR #224, opened 2023-08-16
perftest: support set flow_label through env variable FLOW_LABEL

changchengx:
- What
Add user parameter ```--flow_label=<value>``` to be set by user to affect the udp sport calculation.
- Why
When some customer report the problem while testing under 1) case one: RoCE LAG 2) case two: not back-to-back connected, the udp sport is randomly selected and make the test result unstable. With adding ```--flow_label=<value>``` to set the grh ```flow_label``` value, it can be used to afftect the udp sport to 1) use same tx port for traffic 2) walk through same traffic path between different round of test when the cards are connected with using switches.

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -544,6 +544,11 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 		printf("      --tclass=<value> ");
 		printf(" Set the Traffic Class in GRH (if GRH is in use)\n");
 
+		if (connection_type != RawEth) {
+			printf("      --flow_label=<value> ");
+			printf(" Set the flow_label in GRH (if GRH is in use)\n");
+		}
+
 		if (cuda_memory_supported()) {
 			printf("      --use_cuda=<cuda device id>");
 			printf(" Use CUDA specific device for GPUDirect RDMA testing\n");
@@ -841,6 +846,7 @@ static void init_perftest_params(struct perftest_parameters *user_param)
 	user_param->mr_per_qp			= 0;
 	user_param->dlid			= 0;
 	user_param->traffic_class		= 0;
+	user_param->flow_label			= 0;
 	user_param->flows			= DEF_FLOWS;
 	user_param->flows_burst			= 1;
 	user_param->perform_warm_up		= 0;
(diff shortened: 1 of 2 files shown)


changchengx:
@HassanKhadour Please help review this PR when you're avaiable.

HassanKhadour:
Hi @changchengx , thanks for your contribution!
Is there any reason why not to add the flow_label as perftest param? also please update the man page and the commit message to include the data in https://github.com/linux-rdma/perftest/pull/224#issue-1853239069

changchengx:
@HassanKhadour Do you prefer to add ```flow_label``` as perftest param instead of using environment variable to set ```flow_label```?

HassanKhadour:
@changchengx yes it would be the preferred option

changchengx:
> @changchengx yes it would be the preferred option
>
@HassanKhadour Feedback later.

changchengx:
> > @changchengx yes it would be the preferred option
> 
> @HassanKhadour Feedback to you later.
>
@HassanKhadour Done

src/perftest_parameters.c
```diff
@@ -2819,7 +2825,11 @@ int parser(struct perftest_parameters *user_param,char *argv[], int argc)
 					use_hl_flag = 0;
 				}
 				if (flow_label_flag) {
-					CHECK_VALUE_NON_NEGATIVE(user_param->flow_label,int,"flow label",not_int_ptr);
+					CHECK_VALUE(user_param->flow_label,int,"flow label",not_int_ptr);
+					if (user_param->connection_type == RawEth && user_param->flow_label < 0) {
+						fprintf(stderr," flow label must be non-negative for RawEth\n");
+						return 1;
```
HassanKhadour, line 2831: add free(duplicates_checker) and lets replace "return 1" by enum "return FAILURE"

HassanKhadour:
> > > @changchengx yes it would be the preferred option
> > 
> > 
> > @HassanKhadour Feedback to you later.
> 
> @HassanKhadour Done

Thanks!, just a small fix, and please add data to the commit message as you did in https://github.com/linux-rdma/perftest/pull/224#issue-1853239069

src/perftest_parameters.c
changchengx: @HassanKhadour 
What's the meaning of ```add free(duplicates_checker)```?
The original code checks that ```flow_label``` isn't negative value when ```connection_type is RawEth```. So, I kept the original code logic there.

HassanKhadour: I know, this part of code (long options) needs enhancements, planning to do it in this period, so nvm, I'll fix in a latter commit.

HassanKhadour (approved):

changchengx:
@HassanKhadour 
1. used ```return FAILURE```
2. Added commit msg to explain this PR.

src/perftest_parameters.c
```diff
@@ -544,6 +544,11 @@ static void usage(const char *argv0, VerbType verb, TestType tst, int connection
 		printf("      --tclass=<value> ");
 		printf(" Set the Traffic Class in GRH (if GRH is in use)\n");
 
+		if (connection_type != RawEth) {
+			printf("      --flow_label=<value> ");
+			printf(" Set the flow_label in GRH (if GRH is in use)\n");
```
changchengx, line 549: @HassanKhadour 
Thanks for review. 
Updated to be ```if GRH is in use```.

HassanKhadour: Thanks!

HassanKhadour:
@changchengx Thanks for your contribution! merged

PR #227, opened 2023-10-27
Error out if CuDeviceGetByPCIBusId fails

jithinjosepkl:
Update cuda_memory_init to error out if CuDeviceGetByPCIBusId fails. Otherwise, it will silently pick up device 0 (ie, value taken from perftest_parameters which is memset to 0 initially).

### DIFF
--- a/src/cuda_memory.c
+++ b/src/cuda_memory.c
@@ -117,7 +117,7 @@ int cuda_memory_init(struct memory_ctx *ctx) {
 		printf("Finding PCIe BUS %s\n", cuda_ctx->device_bus_id);
 		err = cuDeviceGetByPCIBusId(&cuda_ctx->device_id, cuda_ctx->device_bus_id);
 		if (err != 0) {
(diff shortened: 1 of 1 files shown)


HassanKhadour:
Hi @jithinjosepkl, thanks for your PR.
There is several errors that can be faced when calling "cuDeviceGetByPCIBusId", as far as I understand, not all of them indicates "No such PCI Bus Id" ( some indicates problems with the cuda driver)

jithinjosepkl:
@HassanKhadour, is your concern mainly about the error log? I can make it more generic saying "cuDeviceGetByPCIBusId failed" or so. But, I want to get this fix in so that it wont silently use GPU0 and messes up performance results.

please advise.

jithinjosepkl:
@HassanKhadour, friendly ping on this PR

HassanKhadour:
Thanks @jithinjosepkl  for your PR, that is right, generic error would be the best in this case. Merged!

PR #228, opened 2023-11-01
Fix output of error messages

ddmatsu:
I found some error messages are printed like this:

```
# ib_write_bw -d rocep133s0f0 192.168.125.211
Couldn't connect to 192.168.125.211:18515
Unable to open file descriptor for socket connection Unable to init the socket connection
```

The patches fix the output.


HassanKhadour:
Thanks for the PR!
merged

PR #229, opened 2023-11-20
Update perftest_resources.c

ecjtusbs:
avoid unnecessary memory allocation when  "mr_per_qp"  flag is not set


HassanKhadour:
@ecjtusbs Thanks for your pr, merged!

PR #230, opened 2023-11-22
Add option for write-with-immediate verb for write_lat and write_bw tests

rauteric:
The write-with-immediate test will poll the cq on the receiver side instead of polling the receive buffer.

### DIFF
--- a/README
+++ b/README
@@ -195,6 +195,11 @@ SEND tests (ib_send_lat or ib_send_bw) flags:
   -g, --mcg=<num_of_qps> 		Send messages to multicast group with <num_of_qps> qps attached to it
   -M, --MGID=<multicast_gid>		In multicast, uses <multicast_gid> as the group MGID
 
+WRITE latency (ib_write_lat) flags:
+-----------------------------------
+
+  --write_with_imm				Use write-with-immediate verb instead of write
+
 ATOMIC tests (ib_atomic_lat or ib_atomic_bw) flags: 
 ---------------------------------------------------
 
--- a/man/perftest.1
+++ b/man/perftest.1
@@ -415,6 +415,10 @@ many different options and modes.
 .B --use_ooo
  Use out of order data placement.
  System support required.
+.TP
+.B --write_with_imm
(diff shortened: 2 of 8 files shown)


HassanKhadour:
HI @rauteric, thanks for your PR, its indeed awesome!!
I have few notes.

1. I think creating different method (run_iter_lat_write_with_imm) is a better option, even though it would have code duplication we would eliminate "if (user_param->verb == WRITE_IMM)" check in this part of code that is critical for performance measurements.
2. switch cases as the following (case WRITE_IMM:
                                                    case WRITE: <do smth>) should both verbs have <do smth> which is more readable
3. is (rcnt + size_per_qp <= user_param->iters), line 4470 perftest_resources.c really needed? or its needed only for run_iter_lat_send where we post the maximum number of RRs
4. ctx_set_recv_wqes documenation in header file mentions only SEND verb, I think after your commit it should also mention the write_with_imm verb.
5. can you please make the commit message more informative and add general description about your implementation?
6. will you please extend this verb to include also the bandwidth option?

rauteric:
Thanks @HassanKhadour, added write-with-immediate for bandwidth test and made suggested changes.

rauteric:
> is (rcnt + size_per_qp <= user_param->iters), line 4470 perftest_resources.c really needed? or its needed only for run_iter_lat_send where we post the maximum number of RRs

Actually, after creating the bw test, I now post all the RWRs for WRITE_IMM verb as well to match SEND behavior. So we do need the `size_per_qp` check.

Also, I can't see why this PR is failing build. It builds fine on my machine.

HassanKhadour:
Hi, Thanks for the info and adding this new verb is such a great thing!
Its styling and redundancy found by cppcheck that are treated as errors
image

rauteric:
Thanks @HassanKhadour, resolved the build issues.

rauteric:
Fixed bug in duration mode in write_bw-with-imm, and rebased.

HassanKhadour:

PR #238, opened 2024-01-02
Display 'N/A' for peak BW instead of '0' when it's not available

blochl:
The display of 0 peak BW is confusing, especially when the user does not choose to run without peak-bw explicitly, but for example runs with '-D'. In such cases peak-bw calculatiuon is disabled internally, and the zero result is confusing, and makes the user assume that something went wrong during the run. If the value is not calculated, it shouldn't be numeric.

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -3749,10 +3749,17 @@ static void write_bw_report_to_file(int out_json_fd, struct perftest_parameters 
 	else if (user_param->output == OUTPUT_MR)
 		dprintf(out_json_fd, "msgRate_avg: %lf,\n", msgRate_avg);
 	else if (user_param->raw_qos)
-		dprintf(out_json_fd, REPORT_FMT_QOS_JSON, size, sl, iters, bw_peak, bw_avg, msgRate_avg);
+		if (user_param->noPeak)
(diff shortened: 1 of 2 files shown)


firasj:
This may break existing scripts parsing the output and expecting a floating point number.
Maybe a warning message would suffice?

blochl:
> This may break existing scripts parsing the output and expecting a floating point number. Maybe a warning message would suffice?

Sure. Where would you put the warning message then?

HassanKhadour:
as @firasj said, I think this change is problematic and can damage many automations that parses the output and will cause it to fail.

blochl:
@HassanKhadour, yes, I agree (sorry if my reply above was misunderstood). I genuinely asked where would you suggest to put this warning message?

The output looks like this now:
```
 #bytes     #iterations    BW peak[Gb/sec]    BW average[Gb/sec]   MsgRate[Mpps]    CPU_Util[%]
 65536      186641           0.00               195.69 		   0.373258	    4.17
```
Can you suggest a proper place? Also what to do in the JSON file? A warning message can't be put there...

blochl:
@HassanKhadour, @firasj - thanks for your feedback. A warning was added instead.

HassanKhadour:
@blochl Thanks for your contribution! merged

PR #245, opened 2024-01-22
add support for ROCm6.0 API changes

edgargabriel:
perftest fails to compile with ROCm 6.0 due to some backward incompatible changes (resp. deprecations). This commit ensure support for ROCm 4.x, 5.x and 6.x for perftest

### DIFF
--- a/src/rocm_memory.c
+++ b/src/rocm_memory.c
@@ -8,7 +8,9 @@
 #include <errno.h>
 #include "rocm_memory.h"
 #include <hip/hip_runtime_api.h>
+#if defined HAVE_HIP_HIP_VERSION_H
 #include <hip/hip_version.h>
+#endif
 #include "perftest_parameters.h"
 
 #define ROCM_CHECK(stmt)			\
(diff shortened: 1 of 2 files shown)


src/rocm_memory.c
```diff
@@ -6,8 +6,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <errno.h>
-#include <hip/hip_runtime_api.h>
 #include "rocm_memory.h"
+#include <hip/hip_runtime_api.h>
+#include <hip/hip_version.h>
```
mrgolin, line 11: Wouldn't you like to check for existence of hip/hip_version.h in autoconfig?

src/rocm_memory.c
```diff
@@ -44,9 +45,13 @@ static int init_rocm(int device_id) {
 
 	hipDeviceProp_t prop = {0};
 	ROCM_CHECK(hipGetDeviceProperties(&prop, device_id));
+#if HIP_VERSION >= 60000000
+	printf("Using ROCm Device with ID: %d, Name: %s, PCI Bus ID: 0x%x, GCN Arch: %s\n",
+	       device_id, prop.name, prop.pciBusID, prop.gcnArchName);
```
mrgolin, line 50: nit: IMHO it would be cleaner to use some local variable for gcnArch / gcnArchName and keep a single printf.

edgargabriel: the problem is that one is an integer(gcnArch), the other a string (gcnArchName). If you want I can print the integer into a string to unify the printf statement, would you prefer that solution?

mrgolin: You can also split the print. I'm ok with either way.

edgargabriel:
@mrgolin thank you for your feedback, I think I incorporated all your suggestions/comments. Please let know if this looks.

edgargabriel:
Just wanted to check back whether there are any further comments / review requests for this pr? If possible, we would love to see this merged within a reasonable time frame. due to a customer request.

HassanKhadour:
Hi @edgargabriel, thanks for your contribution, merged!

PR #248, opened 2024-02-14
Fix man page handling for out-of-tree builds

chuckcranor:
Commit 2941850 added man pages for perftest that links to all perftest
applications.   Unfortunately this broke out-of-tree builds.  If you
check out the tree and do:

 ./autogen.sh
 mkdir build
 cd build
 ../configure
 make

it will fail with this error:

ln: failed to create symbolic link 'man/ib_write_bw.1': No such file or directory

because no "man" subdir has been created in the build directory.

Fix by having "configure" create `$(top_builddir)/man` if not present (using `AC_CONFIG_COMMANDS `in configure.ac) and by adding `$(top_builddir)` and `$(top_srcdir)` as needed to Makefile.am. Build/install once again works for both in-tree and out-of-tree builds.


HassanKhadour:

PR #249, opened 2024-02-20
Fix out-of-tree man page build when top_srcdir is an absolute path

chuckcranor:
Commit 248d1b74 only fixed out-of-tree man page builds when top_srcdir is a relative path.  The configure script uses $0 to set top_srcdir, so if configure is run with an absolute path then top_srcdir will be an absolute path too (e.g. "../configure" sets top_srcdir to ".." while "/path/to/configure" sets top_srcdir to "/path/to").

Update the "ln -s" command that sets up man page links in Makefile.am to handle the case where top_srcdir is an absolute path.


HassanKhadour:
Thanks for your PR, merged!

PR #254, opened 2024-03-14
Support RDMA with Tegra integrated GPUs

agirault:
Follow instructions for Tegra support from the official documentation: https://docs.nvidia.com/cuda/gpudirect-rdma/index.html#changing-the-allocator

Only use the cuda driver API instead of the cuda runtime API.

Ignore whitespace to make the diff to review more relevant: https://github.com/linux-rdma/perftest/pull/254/commits/6cdaf1c438fc33a44d0a0ab2a86209e47dc8dc5c?w=1


HassanKhadour:
Thanks for your PR, merged!

PR #255, opened 2024-03-15
Correct build requirement names

agirault:
Use proper debian package names to display when missing packages at buildtime
- librdmacm-devel -> librdmacm-dev
- libibumad -> libibumad-dev
- pciutils-devel -> libpci-dev

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -61,8 +61,8 @@ AC_PROG_RANLIB
 AC_HEADER_STDC
 AC_CHECK_HEADERS([infiniband/verbs.h],,[AC_MSG_ERROR([ibverbs header files not found])])
 AC_CHECK_LIB([ibverbs], [ibv_get_device_list], [], [AC_MSG_ERROR([libibverbs not found])])
-AC_CHECK_LIB([rdmacm], [rdma_create_event_channel], [], AC_MSG_ERROR([librdmacm-devel not found]))
-AC_CHECK_LIB([ibumad], [umad_init], [LIBUMAD=-libumad], AC_MSG_ERROR([libibumad not found]))
+AC_CHECK_LIB([rdmacm], [rdma_create_event_channel], [], AC_MSG_ERROR([librdmacm-dev not found]))
+AC_CHECK_LIB([ibumad], [umad_init], [LIBUMAD=-libumad], AC_MSG_ERROR([libibumad-dev not found]))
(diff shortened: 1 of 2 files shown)


perftest.spec
```diff
@@ -7,8 +7,8 @@ Group:          Productivity/Networking/Diagnostic
 Source:         http://www.openfabrics.org/downloads/%{name}-%{version}.tar.gz
 Url:            http://www.openfabrics.org
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
-BuildRequires:  libibverbs-devel librdmacm-devel libibumad-devel
-BuildRequires:  pciutils-devel
+BuildRequires:  libibverbs-dev librdmacm-dev libibumad-dev
```
HassanKhadour, line 10: spec file should contain the rpm package naming and not the debian ones
which is correct as it was

configure.ac
```diff
@@ -61,8 +61,8 @@ AC_PROG_RANLIB
 AC_HEADER_STDC
 AC_CHECK_HEADERS([infiniband/verbs.h],,[AC_MSG_ERROR([ibverbs header files not found])])
 AC_CHECK_LIB([ibverbs], [ibv_get_device_list], [], [AC_MSG_ERROR([libibverbs not found])])
-AC_CHECK_LIB([rdmacm], [rdma_create_event_channel], [], AC_MSG_ERROR([librdmacm-devel not found]))
-AC_CHECK_LIB([ibumad], [umad_init], [LIBUMAD=-libumad], AC_MSG_ERROR([libibumad not found]))
+AC_CHECK_LIB([rdmacm], [rdma_create_event_channel], [], AC_MSG_ERROR([librdmacm-dev not found]))
```
HassanKhadour, line 64: this change is only mentioning the debian packages which is not correct.
maybe the best way to do is mentioning both (debian and rpm).

for example here:
rdmacm wasn't found, consider installing librdmacm-dev or librdmacm-devel. (this is mentioning both rpm and debian naming)

HassanKhadour (changes requested):
Thanks for your contribution

perftest.spec
agirault: will discuss in other comment

configure.ac
agirault: good point. fine with only updating error messages for including both rpm and deb? or something else?

PR #256, opened 2024-03-15
Ignore generated config files from git

agirault:
Ignore these files marked as untracked:
Screenshot 2024-03-14 at 12 03 33 PM


HassanKhadour:
Thanks for your PR, merged!

PR #257, opened 2024-03-19
Display proper unit for MiB/sec results

agirault:
MB is for Megabytes, perftest is reporting in Mebibyte.

Make it coherent with the documentation when reporting Gbps:
```
... (instead of MiB/sec)
    Note: MiB=2^20 byte, while Gb=10^9 bits. Use these formulas for conversion
```


HassanKhadour:

PR #259, opened 2024-04-09
Fix format of man pages

tzafrir-mellanox:


HassanKhadour:
@tzafrir-mellanox Thanks for your contribution! merged

PR #263, opened 2024-05-24
Perftest: Fix rx_depth check for XRC

hginjgerx:
When users manually specifies --use_srq in perftest command, the rx_depth will be checked. If rx_depth is less than the number of qps, the process will throw an error and exit.

For XRC SEND where SRQ is definitely used, users normally don't need to manually specifies --use_srq, since the use_srq flag will be set to on when parsing the XRC parameters. However, the XRC parameters parsing is after the SRQ rx_depth check. If rx_depth is less than the number of qps in this case, it will miss the check, size_per_qp in ctx_set_recv_wqes() will become 0 and ibv_post_srq_recv() won't be called.

Move the XRC parameters parsing ahead of SRQ rx_depth check and set --use_srq to on in advance so that the rx_depth error can be thrown.

Fixes: 4c774a951b3c ("Added XRC for ib_send_bw test")

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -1124,7 +1124,19 @@ static void force_dependecies(struct perftest_parameters *user_param)
 		exit (1);
 	}
 
-	if (user_param->use_srq && user_param->num_of_qps > user_param->rx_depth) {
+	/* XRC Part */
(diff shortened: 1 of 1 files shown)


sshaulnv:
Thanks for your pull request, merged!

PR #264, opened 2024-05-24
Perftest: Support selecting congestion control algorithms

hginjgerx:
Support configuring congestion control algorithms with hns direct verbs.

    New option:  --congest_type

    Usage example:
    ./ib_send_bw -d hns_0 --congest_type=DCQCN
    ./ib_send_bw -d hns_0 --congest_type=DCQCN 192.168.100.100

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -34,6 +34,9 @@ static const char *portStates[] = {"Nop","Down","Init","Armed","","Active Defer"
 static const char *qp_state[] = {"OFF","ON"};
 static const char *exchange_state[] = {"Ethernet","rdma_cm"};
 static const char *atomicTypesStr[] = {"CMP_AND_SWAP","FETCH_AND_ADD"};
+#ifdef HAVE_HNSDV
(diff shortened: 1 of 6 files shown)


src/perftest_parameters.c
```diff
@@ -34,6 +34,7 @@ static const char *portStates[] = {"Nop","Down","Init","Armed","","Active Defer"
 static const char *qp_state[] = {"OFF","ON"};
 static const char *exchange_state[] = {"Ethernet","rdma_cm"};
 static const char *atomicTypesStr[] = {"CMP_AND_SWAP","FETCH_AND_ADD"};
+static const char *congestStr[] = {"DCQCN","LDCP","HC3","DIP"};
```
sshaulnv, line 37: if libhns is missing then this will lead to "warning: ‘congestStr’ defined but not used "
maybe its better to add ifdef here.

hginjgerx: > if libhns is missing then this will lead to "warning: ‘congestStr’ defined but not used " maybe its better to add ifdef here.

Sure. Will fix this. Thanks!

sshaulnv:
thanks for your contribution, merged!

PR #272, opened 2024-07-24
Perftest: Add support for TD lock-free mode

hginjgerx:
Add support for TD lock-free mode
New option:  --no_lock

        Usage example:
        ./ib_send_bw -d hns_0 --no_lock
        ./ib_send_bw -d hns_0 --no_lock 192.168.100.100

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -389,6 +389,20 @@ if [test $HAVE_HNSDV = yes]; then
 	AC_SUBST([LIBHNS])
 fi
 
+AC_TRY_LINK([#include <infiniband/verbs.h>],
+	[ibv_cq_ex_to_cq], [HAVE_CQ_EX=yes], [HAVE_CQ_EX=no])
+AM_CONDITIONAL([HAVE_CQ_EX], [test "x$HAVE_CQ_EX" = "xyes"])
+if [test $HAVE_CQ_EX = yes]; then
+	AC_DEFINE([HAVE_CQ_EX], [1], [Have CQ EX API support])
+fi
+
+AC_TRY_LINK([#include <infiniband/verbs.h>],
+	[ibv_alloc_td], [HAVE_TD_API=yes], [HAVE_TD_API=no])
+AM_CONDITIONAL([HAVE_TD_API], [test "x$HAVE_TD_API" = "xyes"])
+if [test $HAVE_TD_API = yes]; then
+	AC_DEFINE([HAVE_TD_API], [1], [Have TD API support])
+fi
+
 CFLAGS="-g -Wall -D_GNU_SOURCE -O3 $CFLAGS"
 LDFLAGS="$LDFLAGS"
 LIBS=$LIBS" -lpthread"
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1317,6 +1317,20 @@ int destroy_ctx(struct pingpong_context *ctx,
 	}
 	#endif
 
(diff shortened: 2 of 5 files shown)


src/perftest_resources.c
```diff
@@ -1517,6 +1531,44 @@ int create_reg_cqs(struct pingpong_context *ctx,
...
+			recv_cq_attr.comp_mask = IBV_CQ_INIT_ATTR_MASK_PD;
+		}
+		ctx->recv_cq = ibv_cq_ex_to_cq(ibv_create_cq_ex(ctx->context, &recv_cq_attr));
+		if (!ctx->recv_cq) {
+			fprintf(stderr, "Couldn't create a receiver CQ\n");
+			return FAILURE;
+		}
+	}
+	return SUCCESS;
+#else
```
sshaulnv, line 1571: i think that you can make the creation of the cq's more generic by using the `ibv_cq_init_attr_ex` and `ibv_cq_ex_to_cq()` instead of 'ibv_create_cq()', and then minimize the filtering of the `HAVE_TD_API` to:
```
if (user_param->no_lock) {
		send_cq_attr.parent_domain = ctx->pad;
		send_cq_attr.comp_mask = IBV_CQ_INIT_ATTR_MASK_PD;
	}
```

hginjgerx: We try to ensure compatibility with older versions of rdma-core, where the cq_ex verbs may not be supported yet.
Since the td verbs were introduced later than the cq_ex verbs, we just use `HAVE_TD_API` here.

sshaulnv: I agree.
My suggestion would be to separate this into two `#ifdef `blocks: one for `HAVE_CQ_EX` and a nested one for `HAVE_TD_API`. This approach will provide greater flexibility for future additions to cq_ex features and make the code easier to review for developers who may not be familiar with TD.

hginjgerx: Sure. Thanks for the suggestion

configure.ac
```diff
@@ -389,6 +389,20 @@ if [test $HAVE_HNSDV = yes]; then
 	AC_SUBST([LIBHNS])
 fi
 
+AC_TRY_LINK([#include <infiniband/verbs.h>],
+	[ibv_cq_ex_to_cq], [HAVE_CQ_EX=yes], [HAVE_CQ_EX=no])
+AM_CONDITIONAL([HAVE_CQ_EX], [test "x$HAVE_CQ_EX" = "xyes"])
+if [test $HAVE_CQ_EX = yes]; then
+	AC_DEFINE([HAVE_CQ_EX], [1], [Have CQ EX APT support])
```
sshaulnv, line 396: "Have CQ EX **API*** support"

hginjgerx: Thanks. Fixed.

sshaulnv:

hginjgerx:
We can check the errno of ibv_create_cq_ex(), and if it is EOPNOTSUPP, use ibv_create_cq() as a fallback.
`/**
static inline
struct ibv_cq_ex *ibv_create_cq_ex(struct ibv_context *context,
				   struct ibv_cq_init_attr_ex *cq_attr)
{
	struct verbs_context *vctx = verbs_get_ctx_op(context, create_cq_ex);

	if (!vctx) {
		errno = EOPNOTSUPP;
		return NULL;
	}

	return vctx->create_cq_ex(context, cq_attr);
}
`
 I'll push a bugfix ASAP.

PR #275, opened 2024-08-06
perftest: Set Ack timeout for rdma_cm connection id

svajrave:
Perftest doesn't configure timeout set by user for rdma_cm connections. It can lead to timeout of mad packets for cetain workloads.

Added code to set the ack timeout of QP associated with rdma_cm connection identifier


sshaulnv:

PR #277, opened 2024-08-19
Perftest: Fix TD lock-free mode not working for QP

hginjgerx:
When creating QP in TD lock-free mode, set attr_ex.pd with ctx->pad instead of ctx->pd, otherwise the lock-free won't work.

Fixes: 38d9fcfc3580 ("Perftest: Add support for TD lock-free mode")


sshaulnv:

PR #286, opened 2024-08-30
Perftest: Fix failure in creating cq when create cq ex is not support…

hginjgerx:
This is to fix the issue mentioned  here #272 

The previous TD lock-free patch introduced ibv_create_cq_ex() to perftest, but it breaks the provider devices which do not support this API yet. For these devices calling this API leads to an errno of EOPNOTSUPP. So add a check of errno, and if it is EOPNOTSUPP, use ibv_create_cq() as a fallback.

Fixes: b6f957f6bc6c ("Perftest: Add support for TD lock-free mode")

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1547,6 +1547,8 @@ int create_reg_cqs(struct pingpong_context *ctx,
 	#endif
 	ctx->send_cq = ibv_cq_ex_to_cq(ibv_create_cq_ex(ctx->context, &send_cq_attr));
 	if (!ctx->send_cq) {
+		if (!user_param->no_lock && errno == EOPNOTSUPP)
+			goto cq_ex_not_supported;
 		fprintf(stderr, "Couldn't create CQ\n");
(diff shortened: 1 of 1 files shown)


hginjgerx:
Hi @sshaulnv 
I'd appreciate it if you could help to test this patch on a bnxt_re device because I don't have one. But I did the testing on a hns device though, where I temporarily removed create_cq_ex() API.

sshaulnv:
Hi,
i also dont have one.
@selvintxavier can you please test the fix?

selvintxavier:
Thanks @hginjgerx and @sshaulnv for fixing it. I have quickly verified on bnxt_re device and it works. Will get back in case we see any issues in future. 

Tested-By: Selvin Xavier<selvin.xavier@broadcom.com>

hginjgerx:
Selvin's Tested-by added.

sshaulnv:
@hginjgerx and @selvintxavier, thank you both for the contribution, merged!

PR #288, opened 2024-09-05
Fix qp_timeout argument in the RDMA CM path

raphael-s-norwitz:
This PR fixes ccae5242bbdb2c8642fdc66d949b901ee3b81eaf and sets qp_timeout when making the connection actually used for data transfer.


rolandd:
FWIW, looks good to me.

src/perftest_communication.c
```diff
@@ -2463,6 +2472,14 @@ int rdma_cm_connection_request_handler(struct pingpong_context *ctx,
 		goto error_2;
 	}
 
+	rc = rdma_set_option(ctx->cm_id, RDMA_OPTION_ID,
```
nzmsv, line 2475: Shouldn't this be set on the accepted `cm_id`?

sshaulnv:
Thanks for your contribution!

PR #290, opened 2024-09-20
Perftest supports MLU latency tests with read/send verbs only

hc235280:


sshaulnv:
Thanks for the contribution, please add more detailed explanation in the commit message and align the man page

PR #301, opened 2024-12-11
Use zero size receive buffers for write w/imm

mrgolin:
WRITE w/ immediate RX completions have zero buffer length. Using the test buffer size is not required and can also lead to errors when posting receive buffers larger than the device supports.


sshaulnv:
Sounds reasonable, thanks for your contribution, merged.

PR #304, opened 2024-12-27
fix core dump issue when disabling HAVE_IBV_WR_API

tianx666:
When using rdma-core-v44, and compile perftest with "bash autoconfig; configure --disable-ibv_wr_api; make", then run a test. A core dump will ocur while ctx->dci_stream_id[i]  is not allocted but assigned 0.


sshaulnv:
Hi @tianx666, thanks for your contribution!

This change will not fully resolve the issue, as there are additional instances where `dci_stream_id` is accessed and also `HAVE_IBV_WR_API` is not directly related to it. 
Since this variable is specific to the DC connection type, a more appropriate approach would be to add a check for the selected connection type before accessing it.

tianx666:
Hi，@sshaulnv 
I've made the changes as you suggested and replaced the macro check with an explicit check before access. As for accessing dci_stream_id, all other instances are inside new_post_send, which is controlled by HAVE_IBV_WR_API, so there should be no issues.

sshaulnv:
it looks great.
thanks for your contribution, merged!

PR #307, opened 2025-01-15
debian: Rely on dh_shlibdeps for determining the dependencies and update build dependencies

bdrung:
Rely on `dh_shlibdeps` to determine the dependencies of perftest (by setting `shlibs:Depends`) instead of hard-coding them. Add the missing build dependency on libibumad-dev and libpci-dev. Building the Debian package on Debian unstable results in these dependencies:
```
 Depends: ibverbs-providers (>= 51), libc6 (>= 2.38), libibumad3 (>= 1.3.9), libibverbs1 (>= 34), libpci3 (>= 1:3.13.0), librdmacm1t64 (>= 1.0.19)
```


debian/control
```diff
@@ -2,15 +2,19 @@ Source: perftest
 Section: net
 Priority: extra
 Maintainer: Gil Rockah <gilr@dev.mellanox.co.il>
-Build-Depends: debhelper (>= 8.0.0), libibverbs-dev (>= 1.1.6), librdmacm-dev
+Build-Depends: debhelper (>= 10),
+               libibumad-dev,
+               libibverbs-dev (>= 1.1.6),
+               libpci-dev,
+               librdmacm-dev (>= 1.0.8)
```
sfff, line 9: [Debian](https://salsa.debian.org/hpc-team/perftest/-/blob/master/debian/control?ref_type=heads):
```
Build-Depends: debhelper-compat (= 13),
               libibumad-dev,
               libibverbs-dev (>= 1.1.2),
               librdmacm-dev (>= 1.0.8),
               libpci-dev,
               pciutils
```
[https://git.launchpad.net/ubuntu/+source/perftest/tree/debian/control?h=applied/ubuntu/devel](Ubuntu):
```
Build-Depends: debhelper-compat (= 13),
               libibumad-dev,
               libibverbs-dev (>= 1.1.2),
               librdmacm-dev (>= 1.0.8),
               libpci-dev,
               pciutils
```
Are the minimum versions for libibverbs-dev and debhelper correct?

debian/control
```diff
@@ -2,15 +2,19 @@ Source: perftest
...
 Standards-Version: 3.9.2
```
sfff, line 10: Perhaps it is also necessary to update the `Standards-Version: 3.9.2` field to something more current?

bdrung: Yes, that can be done in a follow-up merge request.

debian/control
bdrung: debhelper 10 is the minimum to successfully build the package. I can bump it even further, but that will drop support for older releases. So before doing so: what releases should this packaging support?

I added requirement for libibverbs-dev >= 1.1.6 because perftest did depend on libibverbs1 >= 1.1.6.

PR #312, opened 2025-02-06
Fixed issue with loss of one payload part due to skip of one address

TamaraBabayan:
The posted_per_qp parameter was incremented too early, so while increasing the address in payload buffer, was skipped one address. If we try to check received payload, we see that one address is skipping and so one payload buffer is skipped.


TamaraBabayan:
In function <increase_loc_addr> there is logic that reuses the same buffer in cyclic way. This logic depends on parameter posted_per_qp which is incremented before sending to this function. Because of that early increment, the cyclic logic is
being done one step earlier and we skip one address. So the solution is to change place of increment.

sshaulnv:
Hi @TamaraBabayan, thanks for you contribution, merged!

PR #316, opened 2025-02-27
Detect when a device is rxe (soft RoCE) 

tzafrir-mellanox:
Added a heuristic to detect when a device is rxe (rdma-core soft-RoCE driver).

First patch is the required change. Second patch is mostly added  as a motivator and I'm not sure I want to keep it here.

Motivation: I wanted to add some self checks to the package I include in Debian. Debian uses a packages testing framework called autopkgtest. The only device I can get to work on a generic machine is rxe. I created a test for it.

Tests ran generally OK (with no errors. I'm not completely sure if I could have better monitoring of results. I still mark them as "superficial"). But some spit an line to standard error about "Device not recognized to implement inline feature. Disabling it", and autopkgtest's default logic is that output to stderr means an error. So I wanted to avoid that to catch real errors.

** Running the test: **
The second commit has the script. By default it runs everything (including "setup" and "shutdown" operations that create and destroy a rxe interface, accordingly. I have not tested it on a machine with a real interface.

If you actually want to test autopkgtest:

    sudo apt install autopkgtest   # maybe also need to install mmdebootstrap
    sudo autopkgtest-build-qemu unstable /var/cache/autopkgtest/vm/unstable.img
    # In the top-level directory of this repository:
    autopkgtest . -- qemu /var/cache/autopkgtest/vm/unstable.img

### DIFF
--- a/src/perftest_parameters.c
+++ b/src/perftest_parameters.c
@@ -1952,6 +1952,14 @@ const int str_link_layer(const char *str)
 		return LINK_FAILURE;
 }
 
+bool is_rxe_heuristic(const struct ibv_context *context)
+{
+	if (context->device == NULL) {
+		return false;
+	}
+	return ! strncmp("rxe", context->device->name, 3);
+}
+
 /******************************************************************************
  *
  ******************************************************************************/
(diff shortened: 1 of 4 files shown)


src/perftest_parameters.c
```diff
@@ -1952,6 +1952,14 @@ const int str_link_layer(const char *str)
 		return LINK_FAILURE;
 }
 
+bool is_rxe_heuristic(const struct ibv_context *context)
+{
+	if (context->device == NULL) {
+		return false;
+	}
+	return ! strncmp("rxe", context->device->name, 3);
```
sshaulnv, line 1960: any soft roce device will have 'rxe' as prefix?

PR #317, opened 2025-03-04
assorted CUDA data_direct fixes

drossetti:
- return an error if running with a binary built with an old CUDA Toolkit, i.e. older than CUDA 12.8
- return an error when using [CUDA compatibility](https://docs.nvidia.com/deploy/cuda-compatibility/) and the GPU kernel mode driver is too old


src/cuda_memory.c
```diff
@@ -207,10 +210,19 @@ int cuda_memory_allocate_buffer(struct memory_ctx *ctx, int alignment, uint64_t
 				*dmabuf_fd = 0;
 				CUmemRangeHandleType cuda_handle_type = CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD;
 
+				if (cuda_ctx->use_data_direct) {
 				#ifdef HAVE_DMABUF_MAPPING_TYPE_PCIE
```
realarnavgoel, line 214: An alternative is to have define macro for even for the older GPU driver - it wouldn't conflict with older GPU driver as well and it will avoid having `ifdef HAVE_DMABUF_MAPPING_TYPE_PCIE` altogether and just simplify this to a runtime only check.

```
#ifndef CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE
#define CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE 0x1
#endif
```

drossetti: It is rarely a good idea to create a backward compatibility layer in that way.
For example, in cuda.h CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE is an enum and here it would be a macro, which is treated differently by the compiler.

configure.ac
```diff
@@ -307,6 +307,7 @@ if [test "$CUDA_H_PATH" ]; then
 	[int x = CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD|CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED;],
 	[CUDA_DMA_BUF_PARAMETERS_SUPPORT=yes], [CUDA_DMA_BUF_PARAMETERS_SUPPORT=no])
 	if [test "x$HAVE_REG_DMABUF_MR" = "xyes"] && [test "x$HAVE_CUDA_CUMEMGETHANDLEFORADDRESSRANGE" = "xyes"] && [test "x$CUDA_DMA_BUF_PARAMETERS_SUPPORT" = "xyes"]; then
+		HAVE_CUDA_DMABUF=yes
```
sshaulnv, line 310: The issue this will solve was fixed by 38acb9d6acc871030e8a14c17d2f15872339e463

sshaulnv:
@drossetti , please remove `HAVE_CUDA_DMABUF=yes` and i will merge the commit.

PR #320, opened 2025-03-26
Add support for DMA-buffers in Cambricon devices

hc235280:
To build with MLU DMA-BUF support, use:
./configure --enable-mlu  --with-mlu=</usr/local/neuware>

To run with MLU DMA-BUF enabled, use:
ib_write_bw  --use_mlu=device id  --use_mlu_dmabuf

### DIFF
--- a/src/mlu_memory.c
+++ b/src/mlu_memory.c
@@ -43,7 +43,7 @@ struct mlu_memory_ctx {
 	int device_id;
 	CNdev cnDevice;
 	CNcontext cnContext;
-
+	bool use_dmabuf;
 };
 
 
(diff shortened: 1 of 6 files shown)


src/mlu_memory.c
```diff
@@ -145,6 +157,7 @@ int mlu_memory_allocate_buffer(struct memory_ctx *ctx, int alignment, uint64_t s
 	CNresult error;
 	size_t buf_size = (size + ACCEL_PAGE_SIZE - 1) & ~(ACCEL_PAGE_SIZE - 1);
 
+	struct mlu_memory_ctx *mlu_ctx = container_of(ctx, struct mlu_memory_ctx, base);
```
sshaulnv, line 160: please move it inside the `#ifdef HAVE_MLU_DMABUF` scope. it wont be used unless `HAVE_MLU_DMABUF` is defined

hc235280: mlu_memory_allocate_buffer is also used when use mlu

sshaulnv: can you please refer me to the line in this function that `mlu_ctx` is used?

hc235280: I understand, thank you

PR #324, opened 2025-04-27
Perftest: Do not align SRQ recv length to MTU for hns

hginjgerx:
HNS RoCE will check whether the length of the recv sge exceeds the range of the corresponding MR, which can cause hns RoCE to report a CQE error.


hginjgerx:
Gentle ping @sshaulnv

PR #325, opened 2025-05-04
Cuda: Use pcie mapping regardless of data direct

dkkranz:
Add a flag to map the CUDA address range via PCIe BAR1 for GPU-direct regardless of the data direct feature.

Reviewed-by: Michael Margolin <mrgolin@amazon.com>
Reviewed-by: Yonatan Nachum <ynachum@amazon.com>

### DIFF
--- a/src/cuda_memory.c
+++ b/src/cuda_memory.c
@@ -40,7 +40,7 @@ struct cuda_memory_ctx {
 	CUdevice cuDevice;
 	CUcontext cuContext;
 	bool use_dmabuf;
-	bool use_data_direct;
+	bool use_pcie_mapping;
 	int driver_version;
 };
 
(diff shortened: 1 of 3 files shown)


sshaulnv:
Hi @dkkranz,
why did you kept `use_data_direct` user_param if the core functionality of his usage `(cu_flags = CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE;`) is now used only if using use_pcie_mapping?

mrgolin:
@sshaulnv I think it's the opposite, ```use_pcie_mapping``` is contained in ```use_data_direct``` functionality and provides only the ```cu_flags = CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE;``` while ```use_data_direct``` also sets ```MLX5DV_REG_DMABUF_ACCESS_DATA_DIRECT``` flag for ```mlx5dv_reg_dmabuf_mr```. See [here](https://github.com/linux-rdma/perftest/pull/325/files#diff-14fb0caf1a7db9f69068c68a0707a280b0398da4426656bdbd744eb8d071e03eR1900).

sshaulnv:
@mrgolin, you are right, my mistake.
PR looks great, i will run some tests with it and update.

sshaulnv:
@dkkranz, @mrgolin, can you please explain the logic of using use_pcie_mapping but keep the MR registeration with as is (ibv_reg_dmabuf_mr)? (without using data_direct)

dkkranz:
In GB200, The Nvidia driver maps the DMA buf address range by default to the C2C attachment and not the PCI BAR that is intended for GPU direct. Running EFA with GPU direct on this platform requires forcing PCIE mapping, and does not require any changes to the EFA MR registration. Please note that the data_direct flag calls the mlx5dv reg MR and cannot be used.

PR #338, opened 2025-08-03
Perftest: Fix address cycling bug in local address increment logic

NitzanLavy:
The increase_loc_addr() function manages circular buffer addressing. One of the input argument is rcnt, which represents the number of post_send/post_receive operations that were called per QP. Based on rcnt, it internally calculates the local address for the next operation. When setting up receive WQEs, this function is being called with the index of the current operation instead of the amount that was already called. This incorrect indexing caused the last WQE in the buffer to incorrectly point to the first WQE's address in the QP. This bug affects the cyclic address calculation used for cache-efficient data placement. The fix ensures proper address cycling by passing the correct index to the function.

Reviewed-by: Firas Jahjah <firasj@amazon.com>
Reviewed-by: Daniel Kranzdorf <dkkranzd@amazon.com>


NitzanLavy:
Thanks!

PR #339, opened 2025-08-03
Unique WR ID

NitzanLavy:
Replacing wr_id usage with qp_index
    Perftest assumes that all WQEs are created with the QP index as their
    wr_id. This creates a situation where WQEs do not have unique IDs.
    Instead, log the completion QP number, and get the QP index from the
    wr_id through a local variable with an appropriate name.

Use unique WR IDs
    While setting send and receive WQEs, assiging to each WQE a unique wr_id
    instead of qp index. This will allow perftest to detect which work
    request got completed.

    The new wr_id format is as follows:
    bytes 0 - 3: WR Index: Index of WR in perftest wr arrays
    bytes 4 - 5: QP Index: Perftest QP index (not QP num) on which the WR is
                 sent or received
    bytes 6 - 7: reserved

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -3458,7 +3458,7 @@ void ctx_set_send_reg_wqes(struct pingpong_context *ctx,
 
 			ctx->wr[i*user_param->post_list + j].sg_list = &ctx->sge_list[i*user_param->post_list + j];
 			ctx->wr[i*user_param->post_list + j].num_sge = MAX_SEND_SGE;
-			ctx->wr[i*user_param->post_list + j].wr_id   = i;
+			ctx->wr[i*user_param->post_list + j].wr_id   = build_wr_id(i * user_param->post_list + j, i);
 
 			if (j == (user_param->post_list - 1)) {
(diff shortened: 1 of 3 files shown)


NitzanLavy:
@sshaulnv could you please take a look and review?

sshaulnv:
@NitzanLavy, sure, will try my best do it by the end of the week

sshaulnv:
@NitzanLavy , thanks for the PR.
The change seems good and necessary, 2 notes:
1. please take a look at perform_warm_up(), when it set `ctx->wr[index].send_flags |= IBV_SEND_SIGNALED` it assume that it is done per qp, so need to adapt the code over there.
2. can you please rebase?

sshaulnv:
@NitzanLavy, sorry, you are right, my mistake.
i tested the change, looks great, please remove the extra `;` and i will merge.

NitzanLavy:
Thanks @sshaulnv, removed the extra ;.

sshaulnv:
merged, thanks for your contribution!

PR #343, opened 2025-09-01
Extended MR fix for providers which don't support extended.

mellowcandle:
Check for the right errno that returned when mr_reg_ex is not implemented by the provider. It tested EPROTONOSUPPORT but it should have tested for EOPNOTSUPP.

Reviewed-by: Yonatan Nachum <ynachum@amazon.com>
Reviewed-by: Yossi Leybovich <sleybo@amazon.com>

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1832,7 +1832,7 @@ static int register_memory_region(struct pingpong_context *ctx,
 
 	mr = register_func(ctx, user_param, qp_index, flags, dmabuf_fd, dmabuf_offset);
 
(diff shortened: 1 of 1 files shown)


sshaulnv:
@mellowcandle, @mrgolin, the error i get (with mlnx devices) is EPROTONOSUPPORT.
i would suggest to change the condition to:
 `if (!mr && (errno == EOPNOTSUPP || errno == EPROTONOSUPPORT) && register_func != register_mr)`

mrgolin:
Yeah, it depends if the error originated in the kernel or in rdma-core. Please see the patch with your suggestion in https://github.com/linux-rdma/perftest/pull/352,

PR #344, opened 2025-09-02
Perftest: Provide clearer error message for specifying QP number in non bw test

Yue-ByteDance:
Original message says that multiple QPs is only available for bw test, but even if user uses `-q 1`, it would still fail and print error, which doesn't match with the message. Change that message to a clearer expression.


sshaulnv:
Merged, thanks for your contribution!

PR #350, opened 2025-09-16
Fix use of cuCtxCreate with Cuda 13.0 and newer

YonatanNachum:
cuCtxCreate function signature was changed from:
CUresult cuCtxCreate(CUcontext* pctx, unsigned int flags, CUdevice dev)

To:
CUresult cuCtxCreate(CUcontext* pctx, CUctxCreateParams* ctxCreateParams, unsigned int flags, CUdevice dev)

in cuda 13.0 which causes perftest compilation to fail. Add support for the new function signature using the CUDA_VER which stores the configured against Cuda version.


sshaulnv:
can you please share the error?
i tried to compile against cuda 13 and it worked.

sshaulnv:
i see.
thanks for the contribution, merged.

PR #352, opened 2025-09-21
Fix MR registration for providers which don't support extended

mrgolin:
Check for the right errno that returned from the missing implementation.

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -1902,7 +1902,8 @@ static int register_memory_region(struct pingpong_context *ctx,
 
 	mr = register_func(ctx, user_param, qp_index, flags, dmabuf_fd, dmabuf_offset);
 
-	if (!mr && errno == EPROTONOSUPPORT && register_func != register_mr) {
+	if (!mr && (errno == EOPNOTSUPP || errno == EPROTONOSUPPORT) &&
+	    register_func != register_mr) {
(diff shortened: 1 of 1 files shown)


src/perftest_resources.c
```diff
@@ -1902,7 +1902,8 @@ static int register_memory_region(struct pingpong_context *ctx,
 
 	mr = register_func(ctx, user_param, qp_index, flags, dmabuf_fd, dmabuf_offset);
 
-	if (!mr && errno == EPROTONOSUPPORT && register_func != register_mr) {
+	if (!mr && (errno == EOPNOTSUPP || errno == EPROTONOSUPPORT) &&
```
sfff, line 1905: As I see it, when the **ex** function is not supported by the device, the return should only be `errno == EOPNOTSUPP`. Do we really need to keep the check for `errno == EPROTONOSUPPORT`, could that value also occur here?

mrgolin: Actually both can be returned, ```EPROTONOSUPPORT``` is being returned when the kernel doesn't support the command or the supplied args (DMA handle in this case) while ```EOPNOTSUPP``` can be returned by low level drivers when they don't support it. Please see the original [PR](https://github.com/linux-rdma/perftest/pull/343).

sshaulnv:
merged, thanks for your contribution!

PR #353, opened 2025-09-24
Fix set max_rd_atomic respecting the local device limit

sfff:
When configuring max_rd_atomic account for our device's constraint in cases where the destination's max_qp_rd_atom (from ibv_devinfo -v) exceeds our local limit.

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -3061,7 +3061,7 @@ static int ctx_modify_qp_to_rts(struct ibv_qp *qp,
 			attr->timeout   = user_param->qp_timeout;
 			attr->retry_cnt = 7;
 			attr->rnr_retry = 7;
-			attr->max_rd_atomic  = dest->out_reads;
+			attr->max_rd_atomic  = MIN(dest->out_reads, my_dest->out_reads);
 			flags |= (IBV_QP_TIMEOUT | IBV_QP_RETRY_CNT | IBV_QP_RNR_RETRY | IBV_QP_MAX_QP_RD_ATOMIC);
 		}
 	}


sfff:
@sshaulnv, Thank you for the review and the improvement suggestion. I agree that the current approach may appear somewhat complex.
That said, the existing output scheme, although not very simple, already provides all the required information:
1. The test header displays the current parameters [print user_param->out_reads](https://github.com/linux-rdma/perftest/blob/4e4f6629904ba07bf81f207b31433cb9dabbe8ab/src/perftest_parameters.c#L4131)
2. After exchanging information with the remote side, the output includes details about `Outstand reads` on both the local (`local address: LID 0000 QPN ... OUT ...`) and remote sides (`remote address: LID 0000 QPN ... OUT ...`) [print remote OUT](https://github.com/linux-rdma/perftest/blob/4e4f6629904ba07bf81f207b31433cb9dabbe8ab/src/perftest_communication.c#L1832)

Would it make sense to rename `"OUT %#04x"` to `"OUTSTAND %#04x"`? Alternatively, is there a better way to improve the clarity of the output without breaking the established convention?

I would like to additionally point out that this patch is intended to prevent an error in cases where the local value (printed in test header) is lower than the value obtained from the remote side.

sfff:
I would like to clarify why I think keeping the current output format unchanged is the least intrusive option for this fix.

The header prints the local requested/configured test parameters. In contrast, the `local address: / remote address:` lines print the exchanged endpoint data. For tests, `OUT` is part of that exchanged endpoint information, not a separate report of the final QP state.

With this patch, the value programmed into `max_rd_atomic` is derived from the two already printed exchanged values:

    effective max_rd_atomic = min(local OUT, remote OUT)

Adding this value into the existing address line would mix two different meanings in one line: endpoint exchange data and negotiated QP configuration. It would also change an established compact output format that may already be consumed by scripts/parsers.

For that reason, I would prefer to keep this patch limited to the functional fix and avoid changing the output format unless you strongly prefer an explicit additional line.

@sshaulnv Does this explanation address your concern about the report output, or would you still prefer a follow-up output change before merge?

PR #364, opened 2025-12-08
Data Validation for Perftest

NitzanLavy:
This patch series adds data validation capability to perftest for write_bw test scenarios, enabling verification of data integrity during RDMA write operations.

Data validation complements perftest's performance testing by providing functional correctness verification. This is valuable for validating new hardware, drivers, or firmware deployments, debugging intermittent data corruption issues under load, and ensuring RDMA configurations work correctly before production use. The feature is disabled by default and only activated via explicit flags, ensuring no impact on standard performance testing workflows.

Three validation modes are supported:
- random: Randomized data generation
- serial: Sequential numeric series with configurable starting value
- pattern: Custom patterns from input file

Usage:
  --data_validation <random|serial|pattern>
  --data_start_value <N>           (for serial mode, default: 0)
  --payload_file_path <file>       (for pattern mode)

Requirements:
- Must use write_bw with --write_with_imm flag
- post_list (-l) must equal tx_depth (-t)
- recv_post_list must equal rx_depth (-r)

The immediate field is used to pass buffer offset hints for validation. Server-side validation compares expected vs actual data and reports mismatches.

Patches:
1. Perftest: Write BW random data validation
2. Perftest: Write BW serial data validation
3. Perftest: Write BW pattern data validation

### DIFF
--- a/man/perftest.1
+++ b/man/perftest.1
@@ -472,10 +472,11 @@ many different options and modes.
  Use write-with-immediate verb instead of write.
  Write tests only.
 .TP
-.B --data_validation=<random|serial>
+.B --data_validation=<random|serial|pattern>
 Perform data validation on transferred packets.
  random: Data is randomized.
  serial: Data is filled with sequential numeric series. Use --data_start_value=<N> to set starting point.
+ pattern: Data is filled using contents of file provided by --payload_file_path=<file>.
 .TP
 .B --data_start_value
 Starting value for serial data validation. Set to 0 by default.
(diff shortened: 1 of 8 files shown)


NitzanLavy:
Reworking this to coexist with the asynchronous data validation that has since merged, following the two-mode split @sshaulnv proposed above — async stays the default, `--data_validation=sync` selects the general-purpose mode.

It depends on the fixes in #407 and won't work until those land, so I'll push the update once it is merged.

PR #382, opened 2026-03-30
Perftest: Add reason print to bl blka and select err

SherrinZhou:
Make info print for descriptive and useful for the users at several places.

### DIFF
--- a/src/perftest_communication.c
+++ b/src/perftest_communication.c
@@ -1294,9 +1294,10 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters 
 	}
 
 	if (event->event != RDMA_CM_EVENT_ESTABLISHED) {
-		fprintf(stderr, "Unexpected CM event bl blka %d\n", event->event);
+		fprintf(stderr, "Unexpected CM event bl blka %s; error: %d.\n",
+		rdma_event_str(event->event), event->status);
 		rdma_ack_cm_event(event);
-                return FAILURE;
+		return FAILURE;
 	}
 
 	if (user_param->connection_type == UD) {
--- a/src/perftest_resources.h
+++ b/src/perftest_resources.h
@@ -72,6 +72,8 @@
 #include <sys/socket.h>
 #include <netdb.h>
 #include <fcntl.h>
+#include <string.h>
+#include <errno.h>
 #include "perftest_parameters.h"
 
 #ifdef HAVE_CUDA
@@ -950,7 +952,8 @@ static __inline int ctx_notify_send_recv_events(struct pingpong_context *ctx)
 	if (select(MAX(ctx->recv_channel->fd,
 		       ctx->send_channel->fd) + 1,
 		   &rfds, NULL, NULL, NULL) == -1) {
-		fprintf(stderr, "Failed to get completion events\n");
+		fprintf(stderr, "Failed to get completion events: %s\n",
+			strerror(errno));
 		return FAILURE;
 	}
 


src/perftest_communication.c
```diff
@@ -1228,9 +1294,10 @@ int rdma_client_connect(struct pingpong_context *ctx,struct perftest_parameters
 	}
 
 	if (event->event != RDMA_CM_EVENT_ESTABLISHED) {
-		fprintf(stderr, "Unexpected CM event bl blka %d\n", event->event);
+		fprintf(stderr, "Unexpected CM event bl blka %s; error: %d.\n",
+		rdma_event_str(event->event), event->status);
```
maxboone, line 1298: Was about to draft up a similar PR - thanks for writing one already, personally I would add the number as well:

```suggestion
		fprintf(stderr, "Unexpected CM event bl blka %s (%d); error: %d.\n",
		rdma_event_str(event->event), event->event, event->status);
```

SherrinZhou: Thanks for the suggestion! But isn't that number just an enum that is better interpreted in human-readable form by the decode helper? I don't personally find it useful in a diagnostic message.

maxboone: It is, but it's convention / feels natural to me to list both. 

I don't think that you'd quickly run into an errno that is not recognized by `rdma_event_str`, but it is done here as well: https://github.com/linux-rdma/perftest/blob/master/src/perftest_communication.c#L2761

```
	sprintf(error_message, "RDMA CM event error:\nEvent: %s; error: %d.\n",
		rdma_event_str(event->event), event->status);
```

Theoretically, you could get an `UNKNOWN EVENT`: https://github.com/linux-rdma/rdma-core/blob/master/librdmacm/cma.c#L2790

```
const char *rdma_event_str(enum rdma_cm_event_type event)
{
	switch (event) {
...
	default:
		return "UNKNOWN EVENT";
	}
}
```

SherrinZhou: The code you quoted contains the number for the status, not the event, no? I think I did keep the status. I agree with keeping status number since that is particularly useful in a case for rdma disconnect events: a timed-out MAD will generate an otherwise normal-looking event, and that timeout message is only carried in the status. I think the current upstream perftest failed to handle that, and that has caused me some headaches here and there :/ Might as well send a patch adressing that as well...

And regarding the event itself, I believe that is ultimately reported by the Linux kernel IB core. So the UNKNOWN EVENT really is a dead sentinel guard from my POV.

maxboone: Eh, looking at my comment again 🤦‍♂️ - the example actually does not print the event->event number. So it's not convention here.

maxboone: Yeah, you're right - I don't know about the "unknown event" being dead weight - isn't it possible to hit that when your librdmacm version is older than the kernel (and the kernel has a new event that your librdmacm version doesn't have yet)?

SherrinZhou: No prob. Even if that is the convention, I do feel like it's a bit redundant to report the same thing twice

SherrinZhou: > Yeah, you're right - I don't know about the "unknown event" being dead weight - isn't it possible to hit that when your librdmacm version is older than the kernel (and the kernel has a new event that your librdmacm version doesn't have yet)?

Not to my knowledge, no. And I believe the kernel guys treat the "don't break userspace" and backward compatibility with great care.

PR #384, opened 2026-04-09
autotools: fix potential shell syntax errors in configure.ac

taraxacum45e9a:
# Description
This PR fixes a potential shell script syntax error in `configure.ac` by adopting a more robust string comparison pattern for the `HAVE_REG_DMABUF_MR` check.

## The Problem
The existing check used a direct variable expansion:
`if test $HAVE_REG_DMABUF_MR = yes; then`

If `$HAVE_REG_DMABUF_MR` is empty or undefined, the shell evaluates the expression as `if test = yes; then`. This causes a **"unary operator expected"** error during the execution of the `./configure` script, potentially halting the build process or leading to incorrect configuration states.

## The Fix
I have updated the conditional to use the defensive `"x$VAR" = "xyes"` idiom. 
* **Quoting:** Handles cases where the variable is empty or contains whitespace.
* **The "x" Prefix:** Prevents the shell from misinterpreting the variable content as a command-line flag (e.g., if the variable contained `-n`).


sshaulnv:
@taraxacum45e9a thanks for your contribution, merged!

PR #385, opened 2026-04-09
write_bw: guard data_validation_destroy in cleanup path

alephpiece:
This fixes a cleanup crash in `write_bw` when running with `--use_rocm` without `--data_validation`.

Previously, `write_bw` called `data_validation_destroy(&ctx)` unconditionally in the normal cleanup path. For memory backends that do not implement data validation, this could reach an uninitialized optional callback and crash during cleanup.


sshaulnv:

PR #387, opened 2026-04-13
Set recv side SGE to null if verb is write with imm

SreevatsaAnantharamu:
In write with imm, the recv side SGE is not used. Setting this to NULL and num_sge to 0.

### DIFF
--- a/src/perftest_resources.c
+++ b/src/perftest_resources.c
@@ -4099,8 +4099,14 @@ int ctx_set_recv_wqes(struct pingpong_context *ctx,struct perftest_parameters *u
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


mrgolin:
This change breaks EFA that currently doesn't support num_sge = 0. What is the purpose of this change?

sshaulnv:
@mrgolin for write with immediate you dont really need the recv SGE, since the immediate data is part of the cqe.
there are verbs implementation that setting the recv side SGE can break them.

but you are right that we cant assume all support num_sge=0.

@SreevatsaAnantharamu can you please focus the condition?

PR #388, opened 2026-04-14
Added --enable-num_sge_zero in configure.ac to allow for using num_sge=0 for Write with IMM if the vendor supports it

SreevatsaAnantharamu:
Added --enable-num_sge_zero in configure.ac to allow for using num_sge=0 for Write with IMM if the vendor supports it.

Following up on the action item in PR: https://github.com/linux-rdma/perftest/pull/387

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -65,6 +65,11 @@ AS_IF([test "x$enable_cq_ex" = "xno"],
       [USE_CQ_EX=no],
         [USE_CQ_EX=yes])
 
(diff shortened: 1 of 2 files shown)


mrgolin:
Looks good to me, thanks! (worth squashing the [simplified it.](https://github.com/linux-rdma/perftest/pull/388/changes/676d6f635c63af501cab357dc97c28aec274726b) commit)

PR #391, opened 2026-04-22
Add vendor-id and change inline size for nebula-matrix rdma nic

lexxmr:
1. Vendor-id add from 3403 to 3412 for NBL_LEONIS, from 3502 to 350A for NBL_DRACO.
2. NBL_LEONIS support inline_size 32.


sshaulnv:
@lexxmr thanks for your contribution, merged!

PR #396, opened 2026-07-07
Perftest: use memfd for hugepage host memory

### LINKED ISSUE #395: --use_hugepages with --mr_per_qp can exhaust kernel.shmmni
Hi,

I found that perftest can fail to create MR when running with
`--use_hugepages` and `--mr_per_qp`.

When --use_hugepages is used, perftest allocates hugepage-backed host memory
with SysV shared memory:

```c
shmget(IPC_PRIVATE, buf_size, SHM_HUGETLB | IPC_CREAT | SHM_R | SHM_W)
```

This is in src/host_memory.c.
Without --mr_per_qp, perftest allocates one large buffer and shares it
between QPs.
With --mr_per_qp, perftest allocates one buffer per QP. When this is combined
with --use_hugepages, it creates one SysV shm segment per QP. For a large QP
count, this exhausts kernel.shmmni even when enough hugepages are available.

```text
Environment:
- kernel: 6.6.0-28.0.0.34.oe2403.x86_64
- RDMA device: hrn5
- kernel.shmmni: 4096
- HugePages_Total: 30000
- HugePages_Free: 30000
```

Server:
```bash
ib_write_bw -d hrn5_0 -s 2 -Q 1 -q 4096 --use_hugepages --run --mr_per_qp
```

Client:
```bash
ib_write_bw -d hrn5_2 -s 2 -Q 1 -q 4096 --use_hugepages --run --mr_per_qp 192.168.100.1
```

Observed behavior:

```
Failed to allocate hugepages. Please configure hugepages
Failed to allocate hugepage region.
failed to create mr
Failed to create MR
Couldn't create IB resources
```

Workaround:

Increasing `kernel.shmmni` avoids the failure:
```bash
sudo sysctl -w kernel.shmmni=<larger_value>
```
This is only a workaround. The failure happens because the current hugepage
allocation path consumes one SysV shm segment per QP when --use_hugepages
and --mr_per_qp are used together.

Expected behavior:
perftest should not fail because of the SysV shm segment count when enough
hugepages are available.
Possible fix:
Use memfd_create(MFD_HUGETLB) + mmap for --use_hugepages instead of SysV shm.
This still uses hugetlb pages, but does not consume one SysV shm ID per QP.
I have sent a PR for this: #396

Hong-L666:
Fixes #395

This changes the `--use_hugepages` host allocation path from SysV shm to a hugetlb memfd mapping. 

The current code uses `shmget(SHM_HUGETLB)` for hugepage-backed host memory. That is fine for the normal path, where perftest allocates one large buffer and shares it between QPs.

With `--mr_per_qp`, perftest allocates one host buffer per QP. If `--use_hugepages` is also set, each buffer allocation creates one SysV shm segment. With a large QP count this can hit `kernel.shmmni`, even when the hugepage pool still has enough free pages.

Using `memfd_create(MFD_HUGETLB)` keeps the allocation backed by hugetlb pages, but avoids consuming one SysV shm id per buffer. The memfd is closed after `mmap()` succeeds, and the mapping is released with `munmap()`.

Testing:

- `git diff --check`
- `./autogen.sh`
- `./configure`
- `make`

Runtime test:

Server:
```bash
ib_write_bw -d hrn5_0 -s 2 -Q 1 -q 4096 --use_hugepages --run --mr_per_qp
```

Client:
```bash
ib_write_bw -d hrn5_2 -s 2 -Q 1 -q 4096 --use_hugepages --run --mr_per_qp 192.168.100.1
```

Result:

The test passed with this patch. The same command failed before this patch
when the SysV shm segment count reached `kernel.shmmni`.

### DIFF
--- a/src/host_memory.c
+++ b/src/host_memory.c
@@ -6,23 +6,23 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 #include <errno.h>
-#include <sys/ipc.h>
-#include <sys/shm.h>
 #include "host_memory.h"
 #include "host_validation.h"
 #include "perftest_parameters.h"
 
 #if !defined(__FreeBSD__)
+#include <sys/syscall.h>
 #include <sys/mman.h>
+#include <sys/vfs.h>
 #endif
 
-#define SHMAT_ADDR (void *)(0x0UL)
-#define SHMAT_FLAGS (0)
-#define SHMAT_INVALID_PTR ((void *)-1)
 
 #define HUGEPAGE_SIZE_2MB  (2ULL * 1024 * 1024)
 #define HUGEPAGE_SIZE_1GB  (1ULL * 1024 * 1024 * 1024)
+
+#define ALIGN_SIZE(size, align) (((size) + (align) - 1) & ~((align) - 1))
 
 /* MAP_HUGE_* flags for specifying huge page size (Linux 3.8+) */
 #ifndef MAP_HUGE_2MB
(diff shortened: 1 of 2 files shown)


Hong-L666:
Hi @sshaulnv , friendly ping on this PR when you have a chance.

This fixes the kernel.shmmni exhaustion that occurs when --use_hugepages
is combined with --mr_per_qp. With --mr_per_qp, perftest allocates and
registers a separate host buffer for each QP. The existing hugepage
allocation path creates one SysV shared-memory segment for each buffer,
so a large number of QPs can exhaust kernel.shmmni. This patch replaces
the SysV shm-based hugepage allocation with hugetlb-backed memfd mappings.

I tested it with 4096 QPs using `ib_write_bw`; the test passed with this patch, while the same setup failed before the change.

I also noticed that the two GitHub Actions workflows are currently marked as “action required”. Could you please approve the workflows and take a look at the PR when convenient?

I’m happy to add more tests, clarify the kernel compatibility or fallback behavior, or revise the implementation based on your feedback. Thanks!

src/host_memory.c
```diff
@@ -32,29 +31,54 @@
...
+
+#ifndef MFD_HUGE_2MB
+#define MFD_HUGE_2MB MAP_HUGE_2MB
+#endif
+
 #if !defined(__FreeBSD__)
+static int create_hugepage_memfd(const char *name)
+{
+#ifdef SYS_memfd_create
+	return syscall(SYS_memfd_create, name, MFD_HUGETLB | MFD_HUGE_2MB);
```
sshaulnv: The old `shmget(SHM_HUGETLB)` call used whatever huge page size the system defaults to, but the new `memfd_create(MFD_HUGETLB | MFD_HUGE_2MB)` hard-selects the 2 MB pool specifically - so on any system where Hugepagesize isn't 2048 kB (default_hugepagesz=1G, or 64 K-granule arm64 where it's 512 MB), `--use_hugepages` now fails against an empty 2 MB pool even though the user's configured huge pages are sitting right there:

```
Failed to mmap hugepage memfd (errno=12: Cannot allocate memory)
Failed to allocate hugepage region.
failed to create mr
Failed to create MR
 Couldn't create IB resources
```

Hong-L666: Thanks for catching this. You're right—the explicit `MFD_HUGE_2MB` changes the existing behavior by selecting the 2 MB pool instead of the system's default HugeTLB pool.
I've updated the implementation to use `MFD_HUGETLB` and align the mapping to the actual HugeTLB page size reported by the memfd. I'll run the build and runtime tests tomorrow and follow up here with the results.

Hong-L666: I have completed the compilation and traffic testing. Could you please review it further and let me know if there are any areas that need improvement?

Hong-L666:
Hi @sshaulnv , gentle follow-up on the updated patch when you have a chance. The review feedback has been addressed, and the build and runtime tests have passed.

Please let me know if any further changes are needed. Thanks!

sshaulnv:
Hi @Hong-L666 , sorry for the delay, will review this week.

sshaulnv:
@Hong-L666 , thanks for your contribution, merged!

